PR #33, opened 2015-03-04
Set the app environment using the APPSIGNAL_APP_ENV env var

andrielfn:
My current context for this patch:

I have an app running in three environments (qa, stage, production), but I keep only the production file environment, it means that each one runs as "production" (RAILS_ENV=production). The solution I found to use AppSignal is to have different app names for each environment (`my-app-qa`, `my-app-stage`, `my-app-prod`) using the `APPSIGNAL_APP_NAME`.

With this patch, we'll be able to define just one name and differentiate the environment for each app instance (basically is exactly what the `appsignal.yml` does when the app is running in different environments).

I've tried to add a test for this change, but I couldn't. I'd appreciate any help.

### DIFF
--- a/lib/appsignal/integrations/rails.rb
+++ b/lib/appsignal/integrations/rails.rb
@@ -22,7 +22,7 @@ if defined?(::Rails)
           # Load config
           Appsignal.config = Appsignal::Config.new(
             Rails.root,
-            Rails.env,
+            ENV.fetch('APPSIGNAL_APP_ENV', Rails.env),
             :name => Rails.application.class.parent_name
           )
 


matsimitsu:
Hi @andrielfn I'm going to write a spec for it tomorrow, after that we'll be able to merge into master.

Thanks for the contribution!

andrielfn:
@matsimitsu Nice! Let me know if I can help further.

Thank you.

matsimitsu:
Merged! Here are the specs in case you're interested: https://github.com/appsignal/appsignal/commit/c321c6a5d5d0799afd9c15f3b4dbdd9d132cdd95

andrielfn:
@matsimitsu Cool, you have created a new Rails app (`MyEnvApp`). My difficulty was in re-initialize the `MyApp` after set the ENV var. Didn't think about create a new App. :+1: 

Thank you so much!

PR #37, opened 2015-04-20
Notifying via server

naps62:
This change is a result of me trying to understand why none of my apps showed any deploys on AppSignal. It turns out, the current capistrano task notifies the AppSignal service via the same machine that is issuing the deploy.

This is a problem for me (and I would assume for more people) because it requires me to setup the appsignal deploy key both locally as well as in our CI server.
Since we already have that same key set up in production, I see no point in duplicating that

Plus, requiring the key to exist locally (or in the CI) can lead to other errors. For example, one of my coworkers might not have it, which will make his deploy go unnoticed to AppSignal

This proposal makes the deploy notification occur on the server instead, by issuing `appsignal notify_of_deploy` with the appropriate parameters
This way, the server will handle the notification, regardless of who issued the deploy, and there's no need to store the push key outside the server itself

I also added some configurables values, so that other developers can customize the command required to access the appsignal binary. For example, in my case, I had to add `set :appsignal_bin, %i(foreman run appsignal)` since I'm using foreman to load the environment variables in production

### DIFF
--- a/lib/appsignal/integrations/capistrano/appsignal.cap
+++ b/lib/appsignal/integrations/capistrano/appsignal.cap
@@ -1,26 +1,23 @@
 namespace :appsignal do
   task :deploy do
-    env = fetch(:rails_env, fetch(:rack_env, 'production'))
-    user = ENV['USER'] || ENV['USERNAME']
-    revision = fetch(:appsignal_revision, fetch(:current_revision))
-    logger = fetch(:logger, Logger.new($stdout))
+    params = {
+      revision: fetch(:appsignal_revision, fetch(:current_revision)),
+      user: ENV['USER'] || ENV['USERNAME'],
+      environment: fetch(:rails_env, fetch(:rack_env, 'production'))
+    }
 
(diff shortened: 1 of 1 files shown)


naps62:
PS: I noticed that the tests are not passing. I didn't really bother fixing them already, since I would like your input on whether or not this is something you guys are willing to change

naps62:
@thijsc can I get some feedback on this, to know if it's worth it to spend some time making it green? I also have a few doubts about how to go about it

PS: Sorry if you're not the person to contact. I don't know who that would be

thijsc:
Sorry for the delay in responding to this. We were having a discussion about this internally but then forgot to update here.

I think this is a great idea. It is a change that we'll have to test and make part of the 0.12 release since it does change the behavior of this feature. I'll leave this open and merge it when we're ready to move 0.12 to beta.

thijsc:
We've released 1.0 and are now ready to start making changes to the gem again, sorry for the delay.

I'm wondering: Would this mean that the marker gets created from every application server you deploy to? Or just a single one?

PR #40, opened 2015-05-05
added ipc client enqueue retry on drbconnerror

serioushaircut:


thijsc:
Isn't this the same fix as #39? :-)

PR #43, opened 2015-07-07
Custom instrumentation methods opt-in for Object

timonv:
I've been meaning to PR this for some time.

It allows measuring of arbitrary methods without much code overhead. We use it in production a lot.

Not sure if monkey patching Object (even if opt-in) is such a good idea for everyone. Happy to make it a module instead if so desired.


thijsc:
Thanks for the patch @timonv :-)

We're working on making a few changes in this area. I will definitely  draw inspiration from this one :-).

PR #44, opened 2015-07-09
Add Rack streaming listener

matsimitsu:
This listener wraps the rack response in a stream wrapper, so we can catch exceptions in long-running streams.


thijsc:
Closing this one, will open a new one to `develop`

PR #47, opened 2015-07-16
Fix issue with sinatra listener trying to add an exception to the tra…

matsimitsu:
…nsaction when AppSignal is not active


thijsc:
:+1:

PR #52, opened 2015-07-24
Add Rake monitoring

matsimitsu:
We need to decide wether we want to monitor a rake job (with possibly tons of instrumentation and events), or if we only want to catch errors.

I'm not sure what the value in performance monitoring in rake tasks is, since it's already a background task, mostly used for long running (slow) jobs. 

Thoughts?


thijsc:
:+1:

PR #61, opened 2015-08-24
Support ActiveJob wrapped jobs

lucasmazza:
ActiveJob uses a single worker class (`ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper`)
to run all the enqueued jobs, but it exposes the job name as `wrapped` entry and
we can use it to report the underlying job class to AppSignal.

This is a similar approach taken used to display AJ jobs on the the Sidekiq UI
mperham/sidekiq#2259


matsimitsu:
Awesome, thanks! :+1:

PR #63, opened 2015-09-04
Route `add_exception` through Appsignal instead of setting it on the …

matsimitsu:
…transaction directly. This way the active check and the ignored_exception check is done.


thijsc:
:+1:

PR #64, opened 2015-09-09
Develop master merge and big refactor

thijsc:


lib/appsignal/transaction.rb
```diff
@@ -33,68 +34,76 @@ def complete_current!
...
+      @paused = false
+    end
+
+    def paused?
+      @paused == true
     end
 
-    def request
-      @request ||= ::Rack::Request.new(env)
+    def when_finished
```
matsimitsu: What does this do?I can't find any other references to `when_finished`

thijsc: Oh that's a typo :-(. Fixing it, not sure why this doesn't cause a test to fail.

lib/appsignal/transaction.rb
```diff
@@ -33,68 +34,76 @@ def complete_current!
...
+
+      @options = options
+      @options[:params_method] ||= :params
+
+      @transaction_index = Appsignal::Extension.start_transaction(@transaction_id, @namespace)
     end
 
-    def sanitized_environment
-      @sanitized_environment ||= {}
+    def pause!
```
matsimitsu: Is pausing still possible? if not we should remove it or at least log a deprication message.

thijsc: Yes, it's still possible. See the subscriber.

lib/appsignal.rb
```diff
@@ -95,26 +109,35 @@ def listen_for_exception(&block)
...
     end
 
     def add_exception(exception)
-      warn '[DEPRECATION] add_exception is deprecated, use set_exception instead'
+      warn '[DEPRECATION] add_exception is deprecated, use set_error instead'
       set_exception(exception)
     end
 
-    def set_exception(exception)
+    def set_error(exception)
```
matsimitsu: Why do we rename this to `set_error` we never use the term error in the gem anywhere except for here, it might be confusing.

matsimitsu:
I'm not entirely sure about the API exposure of `add_distribution_value` in `appsignal.rb` 

Allthough it does that, it doesn't mean anything if you are working with custom metrics. There we use `measurement` as the name. Maybe we should keep those the same. It would also be nice if it would accept a block and just time for itself.

lib/appsignal.rb
thijsc: Yes we should make a choice here. In the extension and protocol everything is called error, I think I'll revert back to using exception in the Ruby code because that's more in line with how Ruby works.

lib/appsignal/integrations/rake.rb
```diff
@@ -0,0 +1,30 @@
...
+        SecureRandom.uuid,
+        Appsignal::Transaction::BACKGROUND_JOB,
+        Appsignal::Transaction::GenericRequest.new(
+          :params => args
+        )
+      )
+      transaction.set_action(name)
+      transaction.set_error(error)
+      transaction.complete!
+      sleep 0.5 # Give us some time to flush to the agent
```
matsimitsu: Isn't `complete!` blocking untill it's send?

thijsc:
Although I agree with your last comment it's outside of the scope of this pull :-). Opening an issue.

lib/appsignal/integrations/rake.rb
thijsc: No, it only adds it to the queue. Normally that's fine, but if the Rake process terminates immediately after this (which is to be expected) the worker thread will be killed too.

lib/appsignal/rack/rails_instrumentation.rb
```diff
@@ -0,0 +1,41 @@
...
+        )
+        begin
+          @app.call(env)
+        rescue => error
+          transaction.set_error(error)
+          raise error
+        ensure
+          transaction.set_http_or_background_action
+          transaction.set_http_or_background_queue_start
+          transaction.set_metadata('path', request.path)
```
matsimitsu: Is this how metadata works? I thought the key was the "group" and the value a hash of key/values. 

So `transaction.set_metadata('session', request.session)` would generate the session block on an exception view.

lib/appsignal/integrations/rake.rb
matsimitsu: Hmm can we make it blocking by accepting a param? (Not that we should do it right now, I'm just checking if there's a scenario where one would expect an exception, but doesn't get it because the app was terminated too soon.

lib/appsignal/rack/rails_instrumentation.rb
thijsc: No, you're referring to `tranaction.set_error_data`. That accepts json. Metadata is just a string with extra information.

lib/appsignal/integrations/rake.rb
thijsc: That's possible yes.

matsimitsu:
:+1:  Looks good

PR #66, opened 2015-09-10
Error rename

thijsc:


lib/appsignal/transaction.rb
```diff
@@ -99,6 +99,8 @@ def set_metadata(key, value)
 
     def set_error(error)
       return unless error
+      return if Appsignal.is_ignored_error?(error)
```
matsimitsu: This might introduce unexpected behavior, you're now able to send an error regardless if it's ignored or not and we're changing it now.

thijsc: I think this actually makes the behavior of the new version consistent with the old one. We used to have `Appsignal.add_exception` in the integrations, which also does this.

matsimitsu:
:+1:

PR #67, opened 2015-09-10
Make gem configurable using only env vars

thijsc:


matsimitsu:
Nice! Much better :+1:

PR #69, opened 2015-09-15
Improve sql sanitization

thijsc:
Fixed sanitization for escaped quotes within strings. Also rename some constants to make them more clear.

### DIFF
--- a/spec/lib/appsignal/event_formatter/active_record/sql_formatter_spec.rb
+++ b/spec/lib/appsignal/event_formatter/active_record/sql_formatter_spec.rb
@@ -57,14 +57,8 @@ if active_record_present?
 
           it { should == ['Model load', "SELECT `table`.* FROM `table` WHERE `id` = ?"] }
 
-          context "with an escaped single quote" do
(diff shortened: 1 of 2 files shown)


spec/lib/appsignal/event_formatter/active_record/sql_formatter_spec.rb
```diff
@@ -114,12 +114,6 @@
 
             it { should == ['Model load', "\"id\" = ?"] }
           end
-
```
matsimitsu: Why is this spec removed? Is it not a valid case? Is there a new case that should be added?

thijsc: This shouldn't happen in this case because it's a SQL style where the table names are in double quotes and the values in single quotes.

matsimitsu:
So single quotes are never used?

thijsc:
No, double quotes are not used if you use an adapter that uses double quoted table names, postgresql en sqlite in practice.

matsimitsu:
:+1:

Issue #25, opened 2014-11-26
ActiveJobs are all wrapped into one unique report


rnaud:
Hey guys!

On our AppSignal reports, all of our Sidekiq jobs are being reported under an ActiveJob wrapper:
ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper

Any way to fix that?

Arnaud

fwl:
Hello,

Is this still the case or has it been fixed since?

Thank you.

thijsc:
It looks like this issue was caused by the way Sidekiq integrates with ActiveJob. They've now fixed this in 3.3.4:

`Improved ActiveJob integration - Web UI now shows ActiveJobs in a nicer format and job logging shows the actual class name, requires Rails 4.2.2+ [#2248, #2259]`

https://github.com/mperham/sidekiq/blob/master/Changes.md

Could you try that version of Sidekiq?

fwl:
@thijsc I will, thank you for pointing this out!

scottrobertson:
It's doing the same for DelayedJob too

wvengen:
I'm working on a Que adapter and am wondering how to fix this ActiveJob issue. It looks like Sidekiq and DelayedJob integrate with ActiveJob for display purposes, if I'm reading the appsignal source correctly, but how to do that for ActiveJob backends that don't have this feature?

```ruby
cls = args.last['job_class'] if cls == "ActiveJob::QueueAdapters::QueAdapter::JobWrapper"
```

in an integration might work, but I wonder how a more general solution would look like.

Also, ideally I'd like to see the actual ActiveJob info in the action arguments (incl. a mention of the JobWrapper class), but have Appsignal split jobs by ActiveJob's class name. Not sure how that would fit in, though.

matsimitsu:
Hi @wvengen do you have an example/code of your Que adapter? I can take a look and see how best to support AppSigal for it :)

wvengen:
@matsimitsu sure, feel free to participate in #70. Thanks!

thijsc:
We've greatly improved the ActiveJob integration in 1.1, which is now in beta. Could you try that version @rnaud?

PR #70, opened 2015-09-19
Add Que integration

wvengen:
Adds support for the [Que](https://github.com/chanks/que) jobs backend. I haven't this tested yet as much as to be ready for merging, but opening a PR allows us to discuss already.

See also https://github.com/chanks/que/issues/119.

- [x] Report ActiveJob jobs by their original name (#25)
- [x] Make it work on Ruby 1.9.3, using `alias_method` instead of `Module#prepend`.
- [x] Test in production

### DIFF
--- a/lib/appsignal.rb
+++ b/lib/appsignal.rb
@@ -20,6 +20,7 @@ module Appsignal
       require 'appsignal/integrations/sidekiq'
       require 'appsignal/integrations/resque'
       require 'appsignal/integrations/sequel'
+      require 'appsignal/integrations/que'
     end
 
     def load_instrumentations
--- a/lib/appsignal/integrations/que.rb
+++ b/lib/appsignal/integrations/que.rb
@@ -1,0 +1,29 @@
+if defined?(::Que)
+  Appsignal.logger.info('Loading Que integration')
+
+  module Appsignal
+    module Integrations
+      module QuePlugin
+        def _run
+          Appsignal.monitor_transaction(
+            'perform_job.que',
+            :class    => attrs[:job_class],
+            :method   => 'run',
+            :metadata => {
+              :id        => attrs[:job_id],
+              :queue     => attrs[:queue],
+              :run_at    => attrs[:run_at].to_s,
+              :priority  => attrs[:priority],
+              :attempts  => attrs[:error_count].to_i
+            },
+            :params   => attrs[:args]
+          ) { super }
+        end
+      end
+    end
+  end
+
+  class Que::Job
+    prepend Appsignal::Integrations::QuePlugin
+  end
+end
--- a/lib/appsignal/integrations/que.rb
+++ b/lib/appsignal/integrations/que.rb
@@ -8,9 +8,11 @@ if defined?(::Que)
           base.class_eval do
 
             def _run_with_appsignal
+              cls = attrs[:job_class]
+              cls = attrs[:args].last['job_class'] if cls == "ActiveJob::QueueAdapters::QueAdapter::JobWrapper"
               Appsignal.monitor_transaction(
                 'perform_job.que',
-                :class    => attrs[:job_class],
+                :class    => cls,
                 :method   => 'run',
                 :metadata => {
                   :id        => attrs[:job_id],
--- a/lib/appsignal/integrations/que.rb
+++ b/lib/appsignal/integrations/que.rb
@@ -4,26 +4,36 @@ if defined?(::Que)
   module Appsignal
     module Integrations
       module QuePlugin
-        def _run
-          Appsignal.monitor_transaction(
-            'perform_job.que',
-            :class    => attrs[:job_class],
-            :method   => 'run',
-            :metadata => {
-              :id        => attrs[:job_id],
-              :queue     => attrs[:queue],
-              :run_at    => attrs[:run_at].to_s,
-              :priority  => attrs[:priority],
-              :attempts  => attrs[:error_count].to_i
-            },
-            :params   => attrs[:args]
-          ) { super }
+        def self.included(base)
+          base.class_eval do
+
+            def _run_with_appsignal
+              Appsignal.monitor_transaction(
+                'perform_job.que',
+                :class    => attrs[:job_class],
+                :method   => 'run',
+                :metadata => {
+                  :id        => attrs[:job_id],
+                  :queue     => attrs[:queue],
+                  :run_at    => attrs[:run_at].to_s,
+                  :priority  => attrs[:priority],
+                  :attempts  => attrs[:error_count].to_i
+                },
+                :params => attrs[:args]
+              ) { _run_without_appsignal }
+            end
+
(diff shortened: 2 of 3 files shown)


wvengen:
@saturnflyer I suppose the Appsignal-way is to filter events with [post-processing middleware](https://github.com/appsignal/appsignal#postprocessing-middleware).

saturnflyer:
@wvengen that's filtering, yes, but that's not the same as the ability to not monitor.

thijsc:
Our standard way to ignore certain jobs is to add a list of jobs you want to ignore to your config:

```
ignore_actions:
  - Job#method
```

In the `0.12` series which is in beta now you can also configure this with an env var:

```
export APPSIGNAL_IGNORE_ACTIONS = "action1,action2"
```
 @saturnflyer Are you ok with this way to configure it?

The implementation looks good to me. The only thing we need to add before merging is a gemfile in https://github.com/appsignal/appsignal/tree/master/gemfiles and an entry in https://github.com/appsignal/appsignal/blob/master/Rakefile#L1 so the specs for this will run in our normal test runs.

wvengen:
Thanks @thijsc. What do you think about:
- #25 ActiveJob jobs are all reported as its jobwrapper - would it be useful to add a line [like this](https://github.com/appsignal/appsignal/issues/25#issuecomment-141357181)? If this changes later, and someone has `ignore_actions`, this would suddenly start reporting the jobs. Perhaps not a deal-breaker.
- `Module#prepend` fails on Ruby 1.9.3. Would it be better to use the old `alias_method` for compatibility, or do you consider that to be less important?
- I don't see delayed_job or resque in any gemfile or the rakefile, does que need to be added then?

thijsc:
* Adding the AJ fallback is a great idea.
* I'd go for `alias_method`, we still support 1.9.3.
* Sorry, I was a bit confused there. We made a split between involved and less involved integrations to keep the test suite runtime under control. This is definitively a less involved one.

wvengen:
Thanks, I'll come back to that.

saturnflyer:
This post is relevant reading: http://hawkins.io/2015/03/on-ruby/
But here's the important point (emphasis added):

> I call out the Honeybadger gem specifically because was the most recent time I'd been bit by a seemingly good thing promoted in the community: monkey patching third party code. Now I don't fault Honeybadger for making their product this way. It provides their customers with direct business value: "just require 'honeybadger' and you're done!" I don't agree with this sort of practice. There are hundreds and hundreds of libraries doing the same thing. **Monkey patching is fine, just make it opt in at least**. Provide something like your_library-sinatra or honeybadger/sinatra, but sigh this is not the default behavior. 

I'm suggesting that there be a way that users of the gem be able to opt-in to global tracking or be able to opt-in individually.

wvengen:
@saturnflyer this could be a good point. Do you mean this specifically for this PR's use of prepending `Que::Job#_run`, or for how Appsignal handles integrations in general? Some integrations use hooks provided by background job systems, which is cleaner of course. But I'm not sure if that addresses your issue.

I would expect Appsignal to monitor all relevant things by default, including Que.
What do you propose?

saturnflyer:
I haven't looked deeply into this gem, but I think the same rule applies. If merely using the gem makes a blanket change to system, that could be considered invasive.

Perhaps an option could be to recommend that users of the gem who do not want it to automatically affect every job should add `require: false` in their Gemfile and then add something like `require 'appsignal/integrations/que/plugin'` where that `plugin` file has the code to add an updated `_run` method in the `QuePlugin` module.
The existing `'appsignal/integrations/que'` file could require the plugin and automatically insert it.
By doing that, everyone should be able to continue using the gem as they currently do, and people who might want to opt-out in certain cases will be able to do so.

Here's an example of how I'd expect it to look:

```
# in your Gemfile
gem 'appsignal', require: false # be sure to require the plugin you need in your code

# in your background job
require 'appsignal/integrations/que/plugin'
class Job < Que::Job
  prepend Appsignal::Integrations::QuePlugin
end

# in the appsignal/integrations/que file
if defined?(::Que)
  Appsignal.logger.info('Loading Que integration')
  require 'appsignal/integrations/que/plugin'
  class Que::Job
    prepend Appsignal::Integrations::QuePlugin
  end
end

# in the appsignal/integrations/que/plugin file
  module Appsignal
    module Integrations
      module QuePlugin
        def _run
          Appsignal.monitor_transaction(
            'perform_job.que',
            :class    => attrs[:job_class],
            :method   => 'run',
            :metadata => {
              :id        => attrs[:job_id],
              :queue     => attrs[:queue],
              :run_at    => attrs[:run_at].to_s,
              :priority  => attrs[:priority],
              :attempts  => attrs[:error_count].to_i
            },
            :params   => attrs[:args]
          ) { super }
        end
      end
    end
  end
```

thijsc:
@saturnflyer Our customers usually expect everything to work out of the box, so by default the appsignal gem will try to integrate as much as it can. I do agree with you that it should at least be possible to disable specific integrations before they load. 

We already have the basis for making this configurable in place here: https://github.com/appsignal/appsignal/blob/master/lib/appsignal.rb#L16. We could add a config option for this per integration, I think we should do this for the next version of the gem which is currently in beta.

wvengen:
@saturnflyer thanks for your complete example! It shows how to explicitely include changes into the existing infrastructure.

I love how within the Rails community things 'just work' 80% of the time, without having to configure and add code at different places to add a feature with a gem. Separating the feature and loading it into different files is the key here.

Because I think the current code is inline with how the other integrations are written, I'd suggest to keep the code it as it is, and keep having a clear separation of defining extensions and loading them for later, across all of the integrations.

Thanks again for the excellent example given.

thijsc:
@wvengen: Sorry for the delay in merging this pull. We've refactored the integrations for 1.0 and are now ready to make changes again. This means that some changes need to be made to this pull. Are you interested in doing that? Otherwise I'll do it myself :-).

thijsc:
@wvengen Any thoughts on this?

wvengen:
Thanks for checking back. I'll try to have a look at it in the coming weeks, but feel free to have a go at it yourself. Hope to be able to respond later.

thijsc:
I think it shouldn't be too much work to convert this to the new structure. @wvengen How about I convert it and you then test it properly again using your app which presumably uses Que?

wvengen:
Thanks for coming back on this. As we'll not be using this integration actively anymore, this isn't very practical at the moment. Anyone else having an interest in this, please make it known here!

thijsc:
In that case I'm closing this issue. If there's any more interest in this we're definitely open to adding this integration.

licatajustin:
Our team is definitely interested in adding Que integration into AppSignal -- how can we help?

tmiller:
@thijsc I'm working on adding support for this, do we still need to support ruby 1.9.x?

thijsc:
I'm afraid we do, we have quite a number of customers who are still running 1.9 for some apps. Thanks for working on this!

PR #73, opened 2015-09-27
Finish up installation

thijsc:
Finishing up loose ends for release. This adds an install log and verification of the downloaded archives.

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -1,63 +1,85 @@
+require 'digest'
+require 'logger'
 require 'mkmf'
 require 'fileutils'
 require 'open-uri'
 require 'zlib'
 require 'rubygems/package'
+require 'yaml'
 require File.expand_path('../../lib/appsignal/version.rb', __FILE__)
 
-HOST            = 'd135dj0rjqvssy.cloudfront.net'
-SUPPORTED_ARCHS = %w(x86_64-linux x86_64-darwin)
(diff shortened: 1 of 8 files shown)


ext/extconf.rb
```diff
@@ -1,63 +1,85 @@
...
-            File.open(ext_path(entry.full_name), 'wb') do |f|
-              f.write(entry.read)
-            end
+def install
+  logger.info "Installing appsignal agent for Ruby #{RUBY_VERSION} on #{RUBY_PLATFORM}"
+
+  unless AGENT_CONFIG[:triples].keys.include?(ARCH)
+    installation_failed(
+      "AppSignal currently does not support your system architecture (#{ARCH})." \
+      "Please let us know at support@appsignal.com, we aim to support everything our customers run."
```
wesoudshoorn: is this something we can email automatically?

thijsc: This will run on either a development machine or a laptop, we don't have a way to detect how we can reliably send e-mail for everything. Also I'm not sure how people would feel about it.

roytomeij:
Not sure about emailing, but it could be a good idea in general to phone home for some failures (maybe just a simple POST to a public url)?

thijsc:
It would indeed be useful to get a ping on failures. We would need to set up infrastructure for it and implement it in in the gem, so I definitively don't want to add this to the scope before release.

matsimitsu:
:+1:

PR #74, opened 2015-10-07
Stopping in worker processes

thijsc:
Call `Appsignal.stop` in the right location for various tools we support.


matsimitsu:
:+1:

PR #75, opened 2015-10-19
Make sure ActiveAdmin notifications dont break the app

matsimitsu:


thijsc:
:+1: 

This also needs to be ported to `develop`.

PR #77, opened 2015-10-20
Fix for trying to create Rack request for wrong env

thijsc:
When using `send_exception` we passed in the env, which can sometimes be incorrect input for a Rack request. Fix this by not using the env at all.


matsimitsu:
Wouldn't passing an empty hash be easier? (then everything works automagically)

thijsc:
Haven't tried that yet. But it will try to get the fullpath from the request as well, which doesn't make sense.

matsimitsu:
:+1:

PR #83, opened 2015-12-10
DelayedJob Mongoid Support

hcatlin:
Currently, Mongo-based systems show #<BSON::ObjectId> instead of anything readable, because of the way that this hash is later printed. By calling #to_s on it (integer, string, BSON object), we'll make sure that the debug output is useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.11.15
+* Improve Sinatra support
+
 # 0.11.14
 * Support ActiveJob wrapped jobs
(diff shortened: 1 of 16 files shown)


roytomeij:
Thanks @hcatlin! It's 10:47pm here, but after some sleep a (well rested) engineer will get back to you on this one :-)

thijsc:
Thanks for the pull! We've moved this code around a bit, I've added the fix in master here: bdcb8c67a5be5777c1d52c4ee0e21cacef2e61a6

PR #86, opened 2016-01-07
Return transaction stub if no current transaction

thijsc:
Return a stub that accepts any method call instead of `nil`. This makes it safe to call methods on the current transaction regardless of wether AppSignal is active.


matsimitsu:
:+1:

PR #87, opened 2016-01-07
Use Mutex.synchronize instead of Thread.exclusive for ruby 2.3 compatibility

pelf:
Ruby 2.3 deprecated `Thread.exclusive` in favor of using `Mutex.synchronize` directly.


thijsc:
Thanks for the pull! I need to research this a bit, will take a better look tomorrow.

pelf:
The build failed for ruby 1.9, not sure if the problem is related to my change or the tests had problems already.

thijsc:
We don't use `Thread.exclusive` anymore in the `develop` branch, which will shortly be released as the next version of the gem. So I don't want to make this change for the `0.11` series, there's not enough benefit to do a full testing cycle of this change.

Thanks a lot for taking the time to make this change though!

PR #90, opened 2016-01-21
Add Elasticsearch Event Formatter

matsimitsu:
Based on the Elasticsearch-rails instrumentation (https://github.com/elastic/elasticsearch-rails/blob/master/elasticsearch-rails/lib/elasticsearch/rails/instrumentation/publishers.rb and https://github.com/elastic/elasticsearch-rails/tree/master/elasticsearch-rails#activesupport-instrumentation)


thijsc:
:+1:

PR #96, opened 2016-02-29
Try to fetch capistrano3 stage to get current environment before tryi…

### LINKED ISSUE #94: Capistrano 3 integration should use `fetch(:stage)`
Hi there,

I'm trying out the `appsignal` gem, and while it works quite smoothly, I'm having an issue with capistrano deployments: appsignal always thinks I'm deploying to production, while I'm actually deploying to staging.

I think the culprit of this is at `lib/appsignal/integrations/capistrano/appsignal.cap:3` :
```
  1 namespace :appsignal do
  2   task :deploy do
  3     env = fetch(:rails_env, fetch(:rack_env, 'production')) #### <=
  4     user = ENV['USER'] || ENV['USERNAME']
  5     revision = fetch(:appsignal_revision, fetch(:current_revision))
```

Appsignal is using `:rails_env` or `:rack_env`, while it's pretty much standard to use `fetch(:stage)`. This will make appsignal use the right environment. Right now, if I `capistrano staging deploy`, a deploy appears in the production environment. By using `fetch(:stage)`, `staging` is used for deployment tracking, as I think it should be. :) 

Thanks a lot in advance!

matsimitsu:
…ng `rails_env` or `rack_env`, fixes #94

### DIFF
--- a/lib/appsignal/integrations/capistrano/appsignal.cap
+++ b/lib/appsignal/integrations/capistrano/appsignal.cap
@@ -1,6 +1,6 @@
 namespace :appsignal do
   task :deploy do
-    env = fetch(:rails_env, fetch(:rack_env, 'production'))
+    env = fetch(:stage, fetch(:rails_env, fetch(:rack_env, 'production')))
     user = ENV['USER'] || ENV['USERNAME']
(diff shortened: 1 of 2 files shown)


thijsc:
:+1:

PR #101, opened 2016-03-30
Let params sanitizer keep unmodified nil, true and false

thijsc:
Brings the behavior in line with the sanitizer in utils.


matsimitsu:
:+1:

PR #103, opened 2016-04-15
Generic Rack instrumentation

### LINKED ISSUE #99: Basic Rack integration broken? Documentation outdated
Hey,

I am currently in the process of setting Appsignal up in a plain Rack application.
Having done this before, I wanted to use the latest appsignal version this time.
This is when I noticed, that your documentation on http://docs.appsignal.com/getting-started/supported-frameworks.html#rack-other is using the old, now unavailable, `Appsignal::Rack::Listener` and `Appsignal::Rack::Instrumentation` middlewares.

By looking at the source code, I quickly found the potential successor `Appsignal::Rack::StreamingListener`, which I added instead of the middlewares named above.

Now I am facing another problem, my request errors are not being sent to Appsignal, despite having a working configuration (`Appsignal.active?` is true and `Appsignal.send_exception` works).
I even added an error catching middleware AFTER the `Appsignal::Rack::StreamingListener` to confirm my errors are passing through it.
Since my application is in no way delivering a streaming response, I suspect that the middleware is not suited or even broken for my use case.

What do you think?

Cheers
Christoph

thijsc:
Add instrumentation middleware for Rack apps that don't use one of the supported frameworks. Fill `env['appsignal.route']` to let AppSignal group actions.

Fixes #99


thijsc:
@matsimitsu Review please :-)

matsimitsu:
👍  Maybe also update the docs and make clear that the route should be unique and not contain params/dynamic content etc. for best results.

PR #106, opened 2016-04-29
Faraday instrumentation formatter

jeromegn:
This adds a Formatter for the `request.faraday` event.

`FaradayMiddleware::Instrumentation` is in the `faraday_middleware` gem and is widely used. This is a first iteration, adding basic info as the title of the event in appsignal.

I'd like to add more information about the request in the body soon enough, but this is a nice minimal start I think.

What are the accepted values for body_format? I see SQL is set to `1`, but I'm not sure what that entails.

### DIFF
--- a/lib/appsignal/event_formatter/faraday/request_formatter.rb
+++ b/lib/appsignal/event_formatter/faraday/request_formatter.rb
@@ -1,0 +1,13 @@
+module Appsignal
+  class EventFormatter
+    module Faraday
+      class RequestFormatter < Appsignal::EventFormatter
+        register 'request.faraday'
+
+        def format(payload)
+          ["#{payload[:method].to_s.upcase} #{payload[:url]}", nil]
+        end
+      end
+    end
+  end
+end
--- a/lib/appsignal/event_formatter/faraday/request_formatter.rb
+++ b/lib/appsignal/event_formatter/faraday/request_formatter.rb
@@ -5,7 +5,12 @@ module Appsignal
         register 'request.faraday'
 
         def format(payload)
-          ["#{payload[:method].to_s.upcase} #{payload[:url]}", nil]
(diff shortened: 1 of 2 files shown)


matsimitsu:
Good idea to add Farday! 

Is it possible to only use the method/protocol/domain in the title? We use it (among other things) to determine the uniqueness of an event. We try to group as many events together as possible, so the `top events` tab gives you a nice overview of domain performance, here's our `net_http` overview for example:

slow_group_events_-_appsignal_-_appsignal

The `body_format` param only has 1 values right now, `SQL_BODY_FORMAT` (SQL). We use this to handle sanitation on the native (rust) side of the gem for performance reasons. 
In this case our SQL Lexer ( https://github.com/appsignal/sql_lexer ) is used to sanitize the SQL is a _lot_ faster then using a Ruby Regex.

jeromegn:
Right. That makes sense. I just made it return the same title, but I added a body with the full URL. That may be useful for differentiating events to the same API. Both in the events list and in the slow events view.

matsimitsu:
Thanks! 👍

thijsc:
Nice work, thanks!

jeromegn:
I'd like to add many more of those. Provided the UI for displaying the details gets a bit better :) (as briefly discussed on Slack)

schmijos:
I stumbled over this in my search to add context to [Her](https://github.com/remi/her) requests and wrote a small blog article about my findings to help others who miss the connections I missed: https://medium.com/@schmijos/better-appsignal-logs-for-her-bf02728f546a

thijsc:
Thanks so much for the post!

You mention that the query params are also logged. Is that correct? Our default formatter only logs the path.

schmijos:
Your right, thank you. I'm wrong there. It's corrected now.

PR #107, opened 2016-05-03
ensure current transaction is nil on completion

alextk:
If current transaction is not set to nil for any reason, all later requests try to send it and fail. This would render the server's worker unusable which is unacceptable.

### DIFF
--- a/lib/appsignal/transaction.rb
+++ b/lib/appsignal/transaction.rb
@@ -37,8 +37,13 @@ module Appsignal
       end
 
       def complete_current!
-        current.complete
-        Thread.current[:appsignal_transaction] = nil
+        begin
+          current.complete
+        rescue Exception => e
+          Appsignal.logger.error("Failed to complete transaction ##{current.transaction_id}. #{e.message}")
+        ensure
+          Thread.current[:appsignal_transaction] = nil
+        end
       end
     end
 


thijsc:
Thanks, this change makes sense. What kind of failure scenario did you encounter?

thijsc:
Ah got it, didn't make the link to the outstanding support issue straight away :-).

thijsc:
Added a test for this in ad483019a4dc261038666a4791d0e68f4036d43b

alextk:
Hi, thanks for such quick response and approval of pull request. 

For the sake of documentation the scenario is as follows: 
1. POST request is made with params that exceed `Rack::Utils.key_space_limit` which causes exception to be thrown. Note that any subsequent access to this request params causes **additional** exception. 
2. `Appsignal::Transaction.complete_current! ` is invoked from rails_instrumentation.rb middleware `ensure` block, and exception is thrown from `Appsignal::Transaction#sample_data` when trying to access request params. This leaves `Thread.current[:appsignal_transaction]` to point to this bad request and not get reset to `nil`
3. Any subsequent request to this worker would throw exception in rails_instrumentation.rb middleware  `ensure` block because `Appsignal::Transaction.complete_current! ` would use bad request from `Thread.current[:appsignal_transaction]`, thus making the worker unusable.

PR #111, opened 2016-06-01
Only add errors in Sinatra env when raise_errors is off

thijsc:


thijsc:
This change has been merged manually.

PR #112, opened 2016-06-14
Add `queue_time` key to Sidekiq instrumentation

lucasmazza:
I noticed that there is a queue time graph for background hosts on the AppSignal UI and the examples of custom instrumentation mentions the `queue_time` (http://docs.appsignal.com/background-monitoring/custom.html) parameter, but that metric isn't collected by default anywhere, so I added it to the Sidekiq hook to collect it by default for all jobs.

### DIFF
--- a/lib/appsignal/hooks/sidekiq.rb
+++ b/lib/appsignal/hooks/sidekiq.rb
@@ -25,7 +25,8 @@ module Appsignal
           :method      => 'perform',
           :metadata    => formatted_metadata(item),
           :params      => params,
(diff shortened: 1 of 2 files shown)


thijsc:
Nice one! This is ready to merge.

We had some test failures in `master`, could you rebase and force push this branch? We should then get a green build.

lucasmazza:
@thijsc all green!

thijsc:
This is now part of `1.2.1.beta.3`.

PR #212, opened 2016-11-22
Gather custom metrics on Unicorn starts/closes/kills

thijsc:
Experiment to see if we think adding metrics to hooks like this works well.


thijsc:
Closing. The customer had a different idea, we'll get back around to this later on.

PR #266, opened 2017-03-15
Shoryuken instrumentation body arg can be a string

matrinox:
If Shoryuken option for `body_parser` is `:text` or `nil` (or just not passed), then `body` will be a `String` and it will crash at `body.reject {...}`

### DIFF
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -2,6 +2,8 @@ module Appsignal
   class Hooks
     # @api private
     class ShoryukenMiddleware
+      EXCLUDE_KEYS = %w(job_class queue_name arguments).freeze
+
       def call(_worker_instance, queue, sqs_msg, body)
         metadata = {
           :queue => queue
@@ -12,10 +14,10 @@ module Appsignal
           :metadata => metadata
         }
 
-        if body.respond_to?(:reject)
-          exclude_keys = [:job_class, :queue_name, :arguments]
-          metadata.merge!(body.reject { |key| exclude_keys.member?(key.to_sym) })
-          options[:class] = body["job_class"] # can it be a symbol? Shoryuken allows custom serialization
+        if body.is_a?(Hash)
+          body = body.each_with_object({}) { |(key, value), object| object[key.to_s] = value }
+          metadata.merge!(body.reject { |key, _value| EXCLUDE_KEYS.member?(key) })
+          options[:class] = body["job_class"]
           options[:params] = body["arguments"] if body.key?("arguments")
         else
           metadata[:body] = body
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -10,7 +10,9 @@ module Appsignal
           :metadata => metadata
         }
 
-        options[:params] = body.is_a?(Hash) ? body : { :params => body }
+        params = body.is_a?(Hash) ? body : { :params => body }
+        truncated_params = format_args(params)
+        options[:param] = truncated_params
 
         if sqs_msg.attributes.key?("SentTimestamp")
           options[:queue_start] = Time.at(sqs_msg.attributes["SentTimestamp"].to_i / 1000)
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -11,8 +11,9 @@ module Appsignal
         }
 
         params = body.is_a?(Hash) ? body : { :params => body }
-        truncated_params = format_args(params)
-        options[:param] = truncated_params
+        config = { filter_parameters: Appsignal.config[:filter_parameters] }
+        truncated_params = Appsignal::Utils::ParamsSanitizer.sanitize(params, config)
+        options[:params] = truncated_params
 
         if sqs_msg.attributes.key?("SentTimestamp")
           options[:queue_start] = Time.at(sqs_msg.attributes["SentTimestamp"].to_i / 1000)
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -6,8 +6,14 @@ module Appsignal
         metadata = {
           :queue => queue
         }
-        exclude_keys = [:job_class, :queue_name, :arguments]
-        metadata.merge!(body.reject { |key| exclude_keys.member?(key.to_sym) })
+
+        if body.is_a?(String)
+          metadata['body_string'] = body
+        else
+          exclude_keys = [:job_class, :queue_name, :arguments]
+          metadata.merge!(body.reject { |key| exclude_keys.member?(key.to_sym) })
+        end
+
         metadata.merge!(sqs_msg.attributes)
 
         options = {
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -2,30 +2,19 @@ module Appsignal
   class Hooks
     # @api private
     class ShoryukenMiddleware
-      EXCLUDE_KEYS = %w(job_class queue_name arguments).freeze
-
-      def call(_worker_instance, queue, sqs_msg, body)
-        metadata = {
-          :queue => queue
-        }
-
+      def call(worker_instance, queue, sqs_msg, body)
+        metadata = { :queue => queue }.merge(sqs_msg.attributes)
         options = {
+          :class => worker_instance.class.name,
           :method => "perform",
           :metadata => metadata
         }
 
-        if body.is_a?(Hash)
-          body = body.each_with_object({}) { |(key, value), object| object[key.to_s] = value }
-          metadata.merge!(body.reject { |key, _value| EXCLUDE_KEYS.member?(key) })
-          options[:class] = body["job_class"]
-          options[:params] = body["arguments"] if body.key?("arguments")
-        else
-          metadata[:body] = body
+        options[:params] = body.is_a?(Hash) ? body : { :params => body }
+
+        if sqs_msg.attributes.key?("SentTimestamp")
+          options[:queue_start] = Time.at(sqs_msg.attributes["SentTimestamp"].to_i / 1000)
         end
-
-        metadata.merge!(sqs_msg.attributes)
-
-        options[:queue_start] = Time.at(sqs_msg.attributes["SentTimestamp"].to_i / 1000) if sqs_msg.attributes.key?("SentTimestamp")
 
         Appsignal.monitor_transaction("perform_job.shoryuken", options) do
           yield
--- a/lib/appsignal/hooks/shoryuken.rb
+++ b/lib/appsignal/hooks/shoryuken.rb
@@ -10,10 +10,9 @@ module Appsignal
           :metadata => metadata
         }
 
-        params = body.is_a?(Hash) ? body : { :params => body }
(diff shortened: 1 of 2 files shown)


matrinox:
Ok, will do.
It's a tough one to find because most people do JSON

matrinox:
How do you run tests? I tried `bundle exec rspec` and `bundle exec rspec spec` but they produce 8 tests:
```bash
$ bundle exec rspec spec --format=documentation
Running specs in 2.3.1 on x86_64-darwin16

Appsignal::AuthCheck
  #perform_with_result
    should give success message
    should give 401 message
    should give an error message
  transmitting
    #perform
      should not transmit any extra data

Appsignal::CLI::Demo
  without configuration
    returns an error
  with configuration
    without environment
      returns an error
    with environment
      calls Appsignal::Demo transmitter
```
When I run the `spec/lib/appsignal/hooks/shoryuken_spec.rb` specs, it's 5 tests that are clearly not a part of the 8:
```bash
$ bundle exec rspec spec/lib/appsignal/hooks/shoryuken_spec.rb --format=documentation
Running specs in 2.3.1 on x86_64-darwin16

Appsignal::Hooks::ShoryukenMiddleware
  with a performance call
    should wrap in a transaction with the correct params
    should handle string bodies (FAILED - 1)
  with an erroring call
    should add the exception to appsignal (FAILED - 2)

Appsignal::Hooks::ShoryukenHook
  with shoryuken
    #dependencies_present?
      should be truthy
  without shoryuken
    #dependencies_present?
      should be falsy
```

Not in the README.md (unless I'm being stupid)

matrinox:
Also I get a failure:
```
1) Appsignal::Hooks::ShoryukenMiddleware with an erroring call should add the exception to appsignal
     Failure/Error: DEFAULT_FAILURE_NOTIFIER = lambda { |failure, _opts| raise failure }
       Exactly one instance should have received the following message(s) but didn't: set_error
```

matrinox:
Regardless, I've added tests and also fixed my code some more

matrinox:
@tombruijn added tests

tombruijn:
@matrinox Thanks! I'll take a better look today or tomorrow

lib/appsignal/hooks/shoryuken.rb
```diff
@@ -5,16 +5,23 @@ def call(_worker_instance, queue, sqs_msg, body)
...
-        metadata.merge!(sqs_msg.attributes)
 
         options = {
-          :class => body["job_class"],
           :method => "perform",
           :metadata => metadata
         }
-        options[:params] = body["arguments"] if body.key?("arguments")
+
+        if body.respond_to?(:reject)
```
tombruijn: maybe `body.is_a?(Hash)` is better here. We call more than just `reject` on body

spec/lib/appsignal/hooks/shoryuken_spec.rb
```diff
@@ -30,14 +33,48 @@
...
+          :body => body.to_json
+        },
+        :queue_start => Time.parse("1976-11-18 0:00:00UTC").utc
+      )
+      Timecop.freeze(Time.parse("01-01-2001 10:01:00UTC")) do
+        Appsignal::Hooks::ShoryukenMiddleware.new.call(worker_instance, queue, sqs_msg, body.to_json) do
+          # nothing
+        end
+      end
+    end
```
tombruijn: could use an empty line after this "end"

spec/lib/appsignal/hooks/shoryuken_spec.rb
```diff
@@ -30,14 +33,48 @@
...
         },
         :params => ["Model", "1"],
         :queue_start => Time.parse("1976-11-18 0:00:00UTC").utc
       )
+      Timecop.freeze(Time.parse("01-01-2001 10:01:00UTC")) do
+        Appsignal::Hooks::ShoryukenMiddleware.new.call(worker_instance, queue, sqs_msg, body) do
+          # nothing
+        end
+      end
     end
```
tombruijn: could use an empty line after this "end"

spec/lib/appsignal/hooks/shoryuken_spec.rb
matrinox: Do you guys have rubocop running in your checks? Also, is that a possible rubocop rule? I would suggest adding it if it is

lib/appsignal/hooks/shoryuken.rb
matrinox: good call

matrinox:
@tombruijn Can you respond to this comment please? https://github.com/appsignal/appsignal-ruby/pull/266#issuecomment-287222163

The test was failing before hand but is not included in the checks here (i.e. travis). That means the test I added for this is not being run. Should I be concerned?

tombruijn:
Hi @matrinox,

Sorry, I had forgotten about this PR 😖. I'll look at it tomorrow and see why the test isn't being run.

About rubocop, I don't think there's a rule for that by default. Maybe rubocop-rspec has a check for it

tombruijn:
@matrinox from what I can tell the Shoryuken test is being run on TravisCI. If I run the test suite locally (without setting a specific gemfile) there are 646 specs with the Shoryuken tests included. On TravisCI the same amount of tests are run (see https://travis-ci.org/appsignal/appsignal-ruby/jobs/218245102). Not sure why a spec failure wasn't triggered on TravisCI then.

Also, I'm not sure why in your case not the entire test suite is run. `bundle exec rspec` should run all specs in the `spec/` directory. If there are only 8 being run, something might be wrong with the RSpec setup on your machine. What happens when you run: `bundle exec rspec spec/**/*_spec.rb`?

Do you have any other questions or is the PR ready for review?

matrinox:
This is when I run `bundle exec rspec`
screen shot 2017-04-05 at 11 20 50 am
And then when I run `bundle exec rspec spec/**/*_spec.rb`
screen shot 2017-04-05 at 11 21 18 am

As for the PR, it's ready for review. I wasn't able to run the full suite locally, just the tests I've changed. So I'm just trusting travis at this point

tombruijn:
I don't really know why it's acting up like that. Normal RSpec behavior, which we haven't changed, should run all the spec files in the `spec/` dir with the pattern I previously posted.

We can also be very explicit and find all the spec files and then send them to rspec as arguments:

```
find ./spec* -name "*_spec.rb" | xargs bundle exec rspec
```

That's the only way I can think of now to get RSpec to run all the spec files on your machine.

---

> As for the PR, it's ready for review.

Alright, great, I'll start reviewing your PR again with a pair of fresh eyes.

tombruijn:
Hi again @matrinox!

Thanks for bringing this problem to our attention. Looks like there was a slightly bigger issue at hand here. I fixed the support for non-Hash values as params, but also rewrote most of the logic of the Shoryuken integration.

I added a commit to this PR: https://github.com/appsignal/appsignal-ruby/pull/266/commits/12d119604f7da1e78120902e49ec7ddf1a099ba2

Try it out and let me know if it works better now :)

Commit message of linked commit:

> Fix the Shoryuken integration. Looks like the initial version was based
> on our Sidekiq integration, but it does not work like that.
> 
> Updated to work against version 1 through 3 with Shoryuken and this was
> the behavior I could determine from the library.
> 
> The original request that brought the problem to our attention was to
> support string (or any other type than a Hash) as a body for the worker.
> This is now supported by wrapping the body in a Hash with the `:params`
> key. Not ideal, but works for now. If we set a hash directly it's
> ignored, because we fetch and filter the params by first wrapping it in
> a `Appsignal::Transaction::GenericRequest` instance. We can't set a
> params value that is not a Hash directly.
> 
> We recommend that users use a
> `body_parser` for their workers that parses the body in an acceptable
> Hash format. The recommended parser is the JSON parser.
> 
> ```
> class MyWorker
>   include Shoryuken::Worker
> 
>   shoryuken_options body_parser: JSON
> 
>   def perform(sqs_msg, body)
>     puts body
>   end
> end
> ```

tombruijn:
Added a small Shoryuken example app to our example repo for other (future) reviewers that want to test this: https://github.com/appsignal/appsignal-examples/tree/shoryuken

matrinox:
you know what? I didn't run `rake install`. I didn't realize that was a part of testing, just development. Anywho, all good now.

I'll leave it to you, @tombruijn. It looks good, seems like a simplification of the existing changes. You know app signal better than I do :)

tombruijn:
Ah.. alright, I didn't think of `rake install`. Good to see it's working now :)

> I'll leave it to you, @tombruijn. It looks good, seems like a simplification of the existing changes. You know app signal better than I do :)

I don't know Shoryuken that well. I'm not a daily user of it and I don't know all the edge cases. I've added your case and reviewed our previous integration code for it, but let us know if you run into something else

The rest of the team can review it now so we can ship this in AppSignal Ruby gem 2.2 soon-ish

lib/appsignal/hooks/shoryuken.rb
```diff
@@ -1,21 +1,19 @@
...
         options = {
-          :class => body["job_class"],
+          :class => worker_instance.class.name,
           :method => "perform",
           :metadata => metadata
         }
-        options[:params] = body["arguments"] if body.key?("arguments")
-        options[:queue_start] = Time.at(sqs_msg.attributes["SentTimestamp"].to_i / 1000) if sqs_msg.attributes.key?("SentTimestamp")
+
+        options[:params] = body.is_a?(Hash) ? body : { :params => body }
```
matsimitsu: Can we wrap the params with a `format_args` call? This truncates the arguments if they are to long, (e.g. if one of the params is a 10mb XML file, we would prefer not to store it on AppSignal :) 

This is how we use it for Sidekiq: https://github.com/appsignal/appsignal-ruby/blob/master/lib/appsignal/hooks/sidekiq.rb#L16

matrinox:
> Ah.. alright, I didn't think of rake install. Good to see it's working now :)

@tombruijn Do you mind checking [this PR](https://github.com/appsignal/appsignal-ruby/pull/279) that updates the README.md to be more clear on this? Thanks

@matsimitsu All done!

tombruijn:
@matrinox I'll look at your PR.

About what @matsimitsu commented. we actually decided on our private slack that we would rework the `format_args` handling, so I'm working on a PR for that. I'll update this PR after that PR is submitted. Thanks for the update though :)

matrinox:
@tombruijn sounds good!

tombruijn:
Submitted the change in #280.
Will update this PR after that's merged.

matrinox:
@tombruijn I took a stab at it https://github.com/appsignal/appsignal-ruby/pull/266/commits/72b8239d68469ace2a5137f84d4b9f01946477f9

tombruijn:
@matrinox that's great! I updated the specs to test for the new behavior. This will fail at the moment, because #280 isn't merged yet and we haven't rebased this PR on it yet. Once that's done the build should be green and we can merge it.

tombruijn:
parameter value truncation removed from this PR as well after PR #280.

Ready for re-review, mostly just removed logic in commit https://github.com/appsignal/appsignal-ruby/pull/266/commits/18e13c56c6728afc3d70584239018e34f070c72b

tombruijn:
Pfew.. finally merged 🎉 
Thanks for the PR @matrinox !
You expect this to be released in version 2.3 of our gem :)

matrinox:
@tombruijn Thanks! It was a great experience contributing to your repo.

tombruijn:
@matrinox  Thanks for your work on this again! This has now been released in AppSignal for Ruby gem 2.3.0 🎉  https://blog.appsignal.com/2017/07/04/gem-2-3.html

PR #379, opened 2018-01-10
Silence backtrace lines with Rails' backtrace cleaner

jeffkreeftmeijer:
As reported via [support](https://app.intercom.io/a/apps/yzor8gyw/respond/inbox/540654/conversations/13024983021) (private Intercom link).

As discussed with @thijsc. We currently only use the filtering part of Rails' backtrace silencer, but it defaults to both filtering (replacing matching strings) and silencing (removing matching lines). This patch removes the second argument to use Rails' default.

### DIFF
--- a/lib/appsignal/transaction.rb
+++ b/lib/appsignal/transaction.rb
@@ -472,7 +472,7 @@ module Appsignal
 
     def cleaned_backtrace(backtrace)
       if defined?(::Rails) && backtrace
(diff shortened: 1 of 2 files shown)


jeffkreeftmeijer:
Closing this, as Rails' default silencers silence too many log lines for our liking (as explained in https://github.com/appsignal/appsignal-ruby/issues/380)

PR #393, opened 2018-03-29
Improve padrino action names

matsimitsu:
Improve Padrino Action name logic

Instead of using the `full_path` we ask the router for it's parsed path.

Old situation: `Padrinotest::Admin:sessions#/accounts/show/1`
New situation: `Padrinotest::Admin:sessions#/accounts/show/:id`

If all else fails, use `"#{settings.name}#unknown"` to indicate that we can't get the action information.

I've also improved the specs by running the requests through the router, instead of reading mocks/stubs.


jeffkreeftmeijer:
[Related thread](https://app.intercom.io/a/apps/yzor8gyw/inbox/inbox/540654/conversations/15448119937) (private Intercom link)

PR #397, opened 2018-04-09
Refactor event_formatter registration code

### LINKED ISSUE #394: Event formatters not registered after AppSignal is started
Currently our formatters register themselves with AppSignal to format certain instrumentation call payloads (for example: https://github.com/appsignal/appsignal-ruby/blob/master/lib/appsignal/event_formatter/action_view/render_formatter.rb#L6)

This makes it seem that you can write your own formatter, add `register` and it should get picked up by AppSignal.

However, in `Appsignal.start` ( https://github.com/appsignal/appsignal-ruby/blob/master/lib/appsignal/event_formatter/action_view/render_formatter.rb#L6 )  we call a method on the EventFormatter that initialises all registered formatters. This has a side effect that all formatters added after `AppSignal.start` is called, won't get initialised and executed. 

Since `Appsignal.start` is called very early on in for example a Rails project, it's almost impossible to add your own event formatter.

We should initialise the formatter automatically when calling register (ensuring it's only initialised once), and not in `Appsignal.start`

matsimitsu:
Improves usability of registering new event_formatters.

The old situation required new formatters to be included
before `Appsignal.start` was called.

The refactored code allows for new event formatters to be added at any time by
moving the initialize code to the `register` call.

Fixes #394

### DIFF
--- a/lib/appsignal.rb
+++ b/lib/appsignal.rb
@@ -121,6 +121,7 @@ module Appsignal
           config.write_to_environment
           Appsignal::Extension.start
           Appsignal::Hooks.load_hooks
+          Appsignal::EventFormatter.initialize_deprecated_formatters
           initialize_extensions
 
           if config[:enable_allocation_tracking] && !Appsignal::System.jruby?
--- a/lib/appsignal.rb
+++ b/lib/appsignal.rb
@@ -121,7 +121,6 @@ module Appsignal
           config.write_to_environment
           Appsignal::Extension.start
           Appsignal::Hooks.load_hooks
-          Appsignal::EventFormatter.initialize_formatters
           initialize_extensions
 
           if config[:enable_allocation_tracking] && !Appsignal::System.jruby?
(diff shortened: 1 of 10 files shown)


matsimitsu:
> I don't understand why we also need to change the implicit register call. Seems to me that that change is not necessary, why is it needed?

Because the `register` call is executed when reading the code. That means that if you put `register` at the top of a file, none of the other content will be loaded. 

Since we initialize the object in the `register` call, this means that we use the `initialize` method from `EventFormatter`, not from the formatter the `register` call is in. 

Long story short, we need to make sure the entire file is loaded, before we initialize it.

thijsc:
On the one hand I'm not a fan of the more verbose api, but can live with that if it makes the thing more robust.

What I do have a problem with is that we're breaking all existing usage of this. It's not deprecated, it straight away doesn't work anymore. I think we should do an actual deprecation and then make the change in 3.0 since it's breaking backwards compatibility.

Is there any way to do this in a backwards compatible way?

PR #660, opened 2020-10-22
Support #start and #finish on instrumenters

jensljungblad:
Hey there. Currently any instrumentation done with `ActiveSupport::Instrumentation` using `#start` and `#finish` rather than `#instrument` doesn't work. Many libraries etc make use of this way of instrumenting, so it would be welcomed if it functioned correctly!

Would something along these lines work? I tested it a bit and it shows up correctly for me, but I don't know much about Appsignal internals, or if this would cause any problems.

### DIFF
--- a/lib/appsignal/hooks/active_support_notifications.rb
+++ b/lib/appsignal/hooks/active_support_notifications.rb
@@ -44,6 +44,36 @@ module Appsignal
               )
             end
           end
+
+          alias start_without_appsignal start
+
+          def start(name, payload = {})
+            # Events that start with a bang are internal to Rails
+            instrument_this = name[0] != BANG
+
+            Appsignal::Transaction.current.start_event if instrument_this
+
+            start_without_appsignal(name, payload)
+          end
+
+          alias finish_without_appsignal finish
+
+          def finish(name, payload = {})
+            finish_without_appsignal(name, payload)
+
+            # Events that start with a bang are internal to Rails
+            instrument_this = name[0] != BANG
+
+            return unless instrument_this
+
+            title, body, body_format = Appsignal::EventFormatter.format(name, payload)
+            Appsignal::Transaction.current.finish_event(
+              name.to_s,
+              title,
+              body,
+              body_format
+            )
+          end
         end
       end
     end


shairyar:
Gem versions being used.

activesupport (= 5.2.4.3)
appsignal (2.11.0)
http (4.4.1)

jensljungblad:
Still waiting for a reply.

thijsc:
Sorry for the slow reply here! I'm taking a look at what would be needed to get this pull or a similar one merged and will get back to you.

thijsc:
I did a bit of research, it does not appear that `start` and `finish` are part of the public notifications API. Can you point me to an example of where this is used?

jensljungblad:
> I did a bit of research, it does not appear that `start` and `finish` are part of the public notifications API. Can you point me to an example of where this is used?

It's used by the HTTP gem's instrumentation for instance: https://github.com/httprb/http/blob/f311a9b5d27042f581b4073c70fb5d2763f31a27/lib/http/features/instrumentation.rb

Also, the Stripe gem uses hook based instrumentation:

```ruby
Stripe::Instrumentation.subscribe(:request_begin) do |request|
  ...
end

Stripe::Instrumentation.subscribe(:request_end) do |request|
  ...
end
```

So to use that with ActiveSupport you can do:

```ruby
Stripe::Instrumentation.subscribe(:request_begin) do |request|
  ActiveSupport::Notifications.instrumenter.start("request.stripe", request: request)
end

Stripe::Instrumentation.subscribe(:request_end) do |request|
  ActiveSupport::Notifications.instrumenter.finish("request.stripe", request: request)
end
```

jensljungblad:
Here's the documentation for ActiveSupport's instrumenter: https://api.rubyonrails.org/classes/ActiveSupport/Notifications/Instrumenter.html

As you can see both `#start` and `#finish` are public and documented methods.

matsimitsu:
Hi @jensljungblad  

I've forked your pull and implemented tests in this pull: https://github.com/appsignal/appsignal-ruby/pull/663 and we just released `2.11.1.beta.1` with the start/stop fix.

Would you be able to test it to see if it does indeed solve your issue, before we release it as a non-beta version? 

Thanks!

jensljungblad:
> Hi @jensljungblad
> 
> I've forked your pull and implemented tests in this pull: #663 and we just released `2.11.1.beta.1` with the start/stop fix.
> 
> Would you be able to test it to see if it does indeed solve your issue, before we release it as a non-beta version?
> 
> Thanks!

Hi! Tested it out just now and it looks correct in the performance trace:

Screenshot 2020-12-04 at 12 49 37

It also appears to show up on the slow events page:

Screenshot 2020-12-04 at 12 51 24

However, if I click in on the event, either from the performance trace or the slow event page, there's nothing there. No event, everything displays 0 response time etc.

matsimitsu:
@jensljungblad thanks! We wil release v2.11.1 with this implementation.

PR #674, opened 2020-12-29
Fix the ERB instantiation for Ruby 3

GRoguelon:
By reviewing the failing tests, I identified that all tests fail because the ERB instantiation. The old version prevented the last argument to be seen as keyword argument. By removing the splat arguments, it works as expected.

Related to #671


thijsc:
Thanks!

PR #727, opened 2021-05-18
Update RedisIntegration to override write instead of process and change event body

alanhala:
This PR is an idea for https://github.com/appsignal/appsignal-ruby/issues/726.

Basically it changes the `RedisIntegration` to override the `#write` method, which is the method that actually executes the command. It also handles the case in which you want to execute a Lua script so that the event created in appsignal shows what you run.

### DIFF
--- a/lib/appsignal/integrations/redis.rb
+++ b/lib/appsignal/integrations/redis.rb
@@ -4,7 +4,7 @@ module Appsignal
   module Integrations
     module RedisIntegration
       def write(command)
-        sanitized_command = command[0] == :eval ? command[1] : command[0].to_s
+        sanitized_command = command[0] == :eval ? command[1] : "#{command[0]}#{" ?" * (command.size - 1)}"
 
         Appsignal.instrument "query.redis", id, sanitized_command do
           super
--- a/lib/appsignal/integrations/redis.rb
+++ b/lib/appsignal/integrations/redis.rb
@@ -3,12 +3,10 @@
 module Appsignal
   module Integrations
     module RedisIntegration
-      def process(commands, &block)
-        sanitized_commands = commands.map do |command, *args|
-          "#{command}#{" ?" * args.size}"
-        end.join("\n")
+      def write(command)
+        sanitized_command = command[0] == :eval ? command[1] : command[0].to_s
 
-        Appsignal.instrument "query.redis", id, sanitized_commands do
+        Appsignal.instrument "query.redis", id, sanitized_command do
           super
         end
       end
--- a/lib/appsignal/integrations/redis.rb
+++ b/lib/appsignal/integrations/redis.rb
@@ -4,7 +4,12 @@ module Appsignal
   module Integrations
     module RedisIntegration
       def write(command)
-        sanitized_command = command[0] == :eval ? command[1] : "#{command[0]}#{" ?" * (command.size - 1)}"
+        sanitized_command =
+          if command[0] == :eval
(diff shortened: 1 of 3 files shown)


lib/appsignal/integrations/redis.rb
```diff
@@ -3,12 +3,10 @@
 module Appsignal
   module Integrations
     module RedisIntegration
-      def process(commands, &block)
-        sanitized_commands = commands.map do |command, *args|
-          "#{command}#{" ?" * args.size}"
-        end.join("\n")
+      def write(command)
+        sanitized_command = command[0] == :eval ? command[1] : command[0].to_s
```
alanhala, line 7: For the example I didn't include the `?` in the command. I guess what we could do is:
- if command is eval, replace all `ARGV[N]` and `KEYS[N]` with `?`
- if it's not an eval, keep it like it was before: `"#{command}#{" ?" * args.size}"`

Do you think it makes sense?

spec/lib/appsignal/hooks/redis_spec.rb
```diff
@@ -36,13 +36,21 @@
...
+                def write(_command)
+                  nil
+                end
+
+                def read
+                  "value"
+                end
+
+                def call(*args)
+                  super
```
alanhala, line 53: need a different way of handling this. I guess I could call directly to write in the test and stub only that method, just like it was done with `process`

lib/appsignal/integrations/redis.rb
alanhala: I kept the old format for the non script commands here. Implemented here https://github.com/appsignal/appsignal-ruby/pull/727/commits/2a3d68c87c823de2cbb05dc95b405c4538543324

shairyar:
Thanks for submitting it, we will take a look and get back to you.

lib/appsignal/integrations/redis.rb
tombruijn: I'm considering keeping the `?` in for keys and arguments for the eval command, so you can see how many keys and arguments were really given to the eval method and to keep it more consistent.

@alanhala do you see a problem with that?

spec/lib/appsignal/hooks/redis_spec.rb
tombruijn: That's okay for now I think. It's not perfect, but we also don't want to run the actual query here.

lib/appsignal/integrations/redis.rb
tombruijn: How it would look with the question marks for arguments in an eval script: 

```ruby
client = Redis.new
client.eval("return { KEYS, ARGV }", ["k1", "k2"], ["a1", "a2"])
client.eval("return { KEYS, ARGV }", :keys => ["k1", "k2"], :argv => ["a1", "a2"])
```

Screen Shot 2021-05-28 at 17 19 59

alanhala: I'm not sure that would look ok if you have a non one-liner script. But it's fine by me. I would only like to see the script that is called.

alanhala: How do you want to show it if it's a non one-liner script? Just by adding a new line and then the question marks?

tombruijn: It would look something like this, with the question marks on the last line after the end of the eval script.
```
Multi line KEYS
statement ARGS
end ? ? ?
```

tombruijn (approved):

PR #841, opened 2022-04-05
`params` is an attribute, not a method

manuelmeurer:


tombruijn (approved):
Thanks @manuelmeurer !

PR #903, opened 2022-12-16
Add Hanami 2 integration

luismiramirez:
Hanami 2 is now supported. This first version adds a Rack middleware that instruments web requests. To load it, require the instrumentation file after booting the Hanami 2 application.

## Installation instructions:

Require the integration file right after requiring the Hanami boot one. By default this is in `config.ru`

```ruby
# config.ru

require "hanami/boot"
require "appsignal/integrations/hanami"

run Hanami.app
```

The Hanami 2 test app used for testing can be found: https://github.com/appsignal/test-setups/tree/add-hanami2

## Pending

- [ ] Docs update

---

## Next steps:

- Instrument view components
- Instrument ROM


lib/appsignal/rack/hanami_instrumentation.rb
```diff
@@ -0,0 +1,55 @@
...
+        begin
+          Appsignal.instrument("process_action.hanami") do
+            @app.call(env)
+          end
+        rescue Exception => error # rubocop:disable Lint/RescueException
+          transaction.set_error(error)
+          raise error
+        ensure
+          transaction.params = request.params.to_h
+          transaction.set_action_if_nil("#{request.request_method} #{request.path}")
```
unflxw, line 46: The action name should have low cardinality -- meaning it can be used to group items. [We already discussed this on Slack, just leaving the comment here so we don't forget]

So the action name would need to be something like `GET /books/:id` or `Actions::Books::Show` -- but not `GET /books/123`.

luismiramirez: Waiting for some feedback here: https://gitter.im/hanami/chat?at=63a06fbc69ce3c5338359379

If we can't find a solution, we should start instrumenting by framework component instead of Rack.

luismiramirez:
Taking another approach. WIP here: https://github.com/appsignal/appsignal-ruby/commit/461bc7899a86f1113ca7c4c6cdc38a51c1058dae

PR #906, opened 2022-12-28
Add Hanami 2 integration

luismiramirez:
Hanami 2 is now supported.
To load it, require the instrumentation
file after booting the Hanami 2 application.

Instead of instrumenting Rack requests, instrument calls to the `call` method in `Hanami::Action` (which receives a Rack request as an argument) so that we can use the action's class name as the action name.

---

### DIFF
--- a/lib/appsignal/cli/install.rb
+++ b/lib/appsignal/cli/install.rb
@@ -61,6 +61,8 @@ module Appsignal
             install_for_padrino(config)
           elsif installed_frameworks.include?(:grape)
             install_for_grape(config)
+          elsif installed_frameworks.include?(:hanami)
+            install_for_hanami(config)
           elsif installed_frameworks.include?(:sinatra)
             install_for_sinatra(config)
           else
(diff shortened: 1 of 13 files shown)


spec/support/helpers/dependency_helper.rb
```diff
@@ -115,6 +119,14 @@ def que_present?
     dependency_present? "que"
   end
 
+  def hanami_present?
+    dependency_present? "hanami"
+  end
+
+  def hanami2_present?
+    ruby_3_0_or_newer? && hanami_present? && Gem.loaded_specs["hanami"].version >= Gem::Version.new("2.0")
```
unflxw, line 127: Do we need/want to check for Ruby >= 3 explicitly? Hanami 2.0 is already only compatible with Ruby 3.0.

unflxw (approved):
🎉

spec/support/helpers/dependency_helper.rb
luismiramirez: This is to avoid earlier versions launching the suite. But maybe it is too defensive.

lib/appsignal/cli/install.rb
```diff
@@ -162,6 +164,24 @@ def install_for_grape(config)
...
 
+        def install_for_hanami(config)
+          puts "Installing for Hanami"
+          config[:name] = required_input("  Enter application name: ")
+          puts
+          configure(config, %w[development production staging], true)
+
+          puts "Finish Hanami installation"
+          puts "  Hanami requires some manual configuration."
+          puts "  After installing the gem, add the following line to config.ru:"
```
unflxw, line 175: Should the instructions say where to add the require line? Like, "after `require hanami/boot`" or "before `Hanami.run`"? Or does it not matter?

unflxw (approved):

PR #922, opened 2023-02-22
improvements/replacing http with https on comments

jotolo:
Adjust some URLs in documentation or comments to use HTTPS instead of HTTP


shairyar:
Thanks @jotolo we will take a look

unflxw:
Thanks @jotolo! I have rebased both your commits into one and fixed the failing test. Many thanks for your contribution. Merging! :D

PR #928, opened 2023-03-23
allow ignoring error hierarchy

jjb:
i just thew this together for discussion, didn't test it

pretty sure you will want to put that `IGNORED_ERROR_LIST` somewhere else 😅


jjb:
in convo with appsignal support, this won't be supported for now

PR #929, opened 2023-03-25
support "warning" alias for WARN log level

jjb:
There are probably people who have "warning" configured but are actually running at "info", and depending on it for alerting, stats, etc, without realizing it, so this might break their instrumentation, so maybe should be in a major version.


shairyar:
https://app.intercom.com/a/inbox/yzor8gyw/inbox/admin/4356044/conversation/16410700181835 (Private link)

tombruijn (approved):

Issue #930, opened 2023-03-27
`Appsignal::Logger` breaks Rails' stdout logger broadcast feature


tbuehlmann:
Learned about Appsignal's new Logging feature today and fiddled with it, exciting! Had it online on a staging environment real quick and it was working, but the rails server wouldn't log to stdout anymore, which was the case before switching things.

Turns out, if you run a rails console or a rails server started using `bin/rails server` (using `--log_to_stdout` or in development), rails will add a broadcast to `Rails.logger` that will let it log to stdout as well as what `Rails.logger` logs to.

Rails basically does this:

```ruby
console = ActiveSupport::Logger.new(STDOUT)
console.formatter = Rails.logger.formatter
console.level = Rails.logger.level

unless ActiveSupport::Logger.logger_outputs_to?(Rails.logger, STDERR, STDOUT)
  Rails.logger.extend(ActiveSupport::Logger.broadcast(console))
end
```

… where `ActiveSupport::Logger.broadcast` returns a module that defines `#add`:

```ruby
Module.new do
  define_method(:add) do |*args, &block|
    logger.add(*args, &block)
    super(*args, &block)
  end

  …
end
```

So `Rails.logger.add(…)` will log to stdout and then call the logger's regular `add` method. **It will only do this if the `#add` method is used**. Now the bug:

It doesn't work with `Appsignal::Logger` as `#info`, `#debug` and the other severity methods **don't internally call `#add`**. Example:

```ruby
def info(message = nil, attributes = {})
  return if INFO < level
  message = yield if message.nil? && block_given?
  return if message.nil?
  Appsignal::Extension.log(
    @group,
    3,
    message,
    Appsignal::Utils::Data.generate(attributes)
  )
end
```

It would be nice if the severity methods would call `#add` internally, too, so the broadcast is working properly. As I don't need any Appsignal attributes in my logs yet, I came up with a custom logger that works for me:

```ruby
# lib/simple_appsignal_logger.rb

class SimpleAppsignalLogger < Logger
  # From https://github.com/appsignal/appsignal-ruby/blob/9f77f6446e2c93ab2b073608fff1135f7229d757/lib/appsignal/logger.rb#L36-L49
  APPSIGNAL_SEVERITIES = Hash.new(0).tap do |severities|
    severities[DEBUG] = 2
    severities[INFO] = 3
    severities[WARN] = 5
    severities[ERROR] = 6
    severities[FATAL] = 7
  end

[...]

PR #938, opened 2023-04-03
Fix Ruby Logger add method signature

luismiramirez:
On the previous release, we added a new attribute to the `add` method of our Logger, which overwrites Ruby's base loggers. This made ActiveSupport Logger broadcast to fail as we called the function with 4 arguments internally after it was overwrittend by ActiveSupport, which expects 3 arguments.

This commits addresses the problem by setting the attributes as an instance variable that is later taken inside the `add` method, so we can keep the signature untouched.

Fixes: https://github.com/appsignal/appsignal-ruby/issues/930


thijsc (approved):

lib/appsignal/logger.rb
```diff
@@ -109,7 +111,20 @@ def fatal(message = nil, attributes = {})
...
+      add_with_attributes(FATAL, message, @group, attributes)
+    end
+
+    private
+
+    def add_with_attributes(severity, message, group, attributes)
+      Thread.current[:appsignal_logger_attributes] = attributes
+      add(severity, message, group)
+    ensure
+      Thread.current[:appsignal_logger_attributes]
```
tbuehlmann, line 123: This line does nothing, probably missing something?

luismiramirez: No, that's correct. The `nil` assignation is missing. Fixed.

Issue #948, opened 2023-04-07
Logging feature


skyporter:
On Rails 6.1 with ruby 3.2.1 and appsignal gem 3.4.0, when I activate the logging feature by adding this code on `config/environment.rb`:

```ruby
Rails.logger = ActiveSupport::TaggedLogging.new(Appsignal::Logger.new("rails"))
```

I get an error message when I request pages:

```
undefined method `silence' for #<Appsignal::Logger:0x00007f11f5ea14c0 @group="rails", @level=1, @format=0, @mutex=#<Thread::Mutex:0x00007f11f5e90f30>, @formatter=#<ActiveSupport::Logger::SimpleFormatter:0x00007f11f5e90dc8 @datetime_format=nil, @thread_key="activesupport_tagged_logging_tags:91120">>
```

(note: this occurs only on staging environment)

Backtrace:
```
activerecord-session_store (2.0.0) lib/action_dispatch/session/active_record_store.rb:119 get_session_model
activerecord-session_store (2.0.0) lib/action_dispatch/session/active_record_store.rb:147 find_session
lib/rack/session/abstract/id.rb:314 load_session
lib/action_dispatch/middleware/session/abstract_store.rb:45 block in load_session
lib/action_dispatch/middleware/session/abstract_store.rb:53 stale_session_check!
lib/action_dispatch/middleware/session/abstract_store.rb:45 load_session
lib/action_dispatch/request/session.rb:234 load!
lib/action_dispatch/request/session.rb:230 load_for_write!
lib/action_dispatch/request/session.rb:132 []=
app/controllers/base_controller.rb:46 set_session_request_info
lib/active_support/callbacks.rb:427 block in make_lambda
lib/active_support/callbacks.rb:198 block (2 levels) in halting
lib/abstract_controller/callbacks.rb:34 block (2 levels) in <module:Callbacks>
lib/active_support/callbacks.rb:199 block in halting
lib/active_support/callbacks.rb:512 block in invoke_before
lib/active_support/callbacks.rb:512 each
lib/active_support/callbacks.rb:512 invoke_before
lib/active_support/callbacks.rb:115 block in run_callbacks
lib/action_text/rendering.rb:20 with_renderer
lib/action_text/engine.rb:59 block (4 levels) in <class:Engine>
lib/active_support/callbacks.rb:126 instance_exec
lib/active_support/callbacks.rb:126 block in run_callbacks
lib/active_support/callbacks.rb:137 run_callbacks
[...]

PR #951, opened 2023-04-12
Fix ActiveSupport::TaggedLogging silence usage

luismiramirez:
When using ActiveSupport::TaggedLogging without the broadcast feature, any call to `silence` fails.

This patch fixes that behaviour by adding the `silence` method to our own logger. It acts as the one defined when using `ActiveSupport`'s broadcast feature. It just calls the passed block and let's the execution to continue.

This has to be done in order to add full support to `ActiveSupport::TaggedLogging`.

Part of: #948 

---

[Maintainer stating that tagged logging is only intended to work with loggers that respond to `silence`.](https://github.com/rails/sprockets-rails/issues/366#issuecomment-229812270)


.changesets/fix-a-bug-when-using-activesupport--taggedlogging.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
+Fix a bug when using ActiveSupport::TaggedLogging
```
tombruijn, line 6:
```suggestion
Fix a bug when using ActiveSupport::TaggedLogging calling the `silence` method.
```

spec/lib/appsignal/logger_spec.rb
```diff
@@ -67,6 +67,18 @@
...
 
+  describe "#silence" do
+    it "calls the given block" do
+      num = 1
+
+      logger.silence do
+        num += 1
+      end
+
+      expect(num).to eq(2)
```
tombruijn, line 78:
```suggestion
      expect(num).to eq(2)
      expect(Appsignal::Extension).to_not receive(:log)
```

Could help to make sure we don't log anything through this method.

tombruijn (approved):
If build green 👍

unflxw (approved):

lib/appsignal/logger.rb
```diff
@@ -114,6 +114,19 @@ def fatal(message = nil, attributes = {})
...
 
+    # When using ActiveSupport::TaggedLogging without the broadcast feature,
+    # the passed logger is required to respond to the `silence` method.
+    # In our case it behaves as the broadcast feature of the Rails logger, but
+    # we don't have to check if the parent logger has the `silence` method defined
+    # as our logger directly inherits from Ruby base logger.
+    #
+    # Links:
+    # https://github.com/rails/rails/blob/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/activesupport/lib/active_support/logger.rb#L60-L76
+    # https://github.com/rails/rails/blob/main/activesupport/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/active_support/logger_silence.rb
```
tombruijn, line 125:
```suggestion
    # rubocop:disable Layout/LineLength
    # https://github.com/rails/rails/blob/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/activesupport/lib/active_support/logger.rb#L60-L76
    # https://github.com/rails/rails/blob/main/activesupport/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/active_support/logger_silence.rb
    # rubocop:enable Layout/LineLength
```
These lines probably fail RuboCop when merged after merging #952.

lib/appsignal/logger.rb
```diff
@@ -114,6 +114,19 @@ def fatal(message = nil, attributes = {})
...
+    def silence(_severity = ERROR, &block)
+      block.call
```
tombruijn, line 127:
```suggestion
    def silence(_severity = ERROR)
      yield
```
Can also use `yield` here directly.

lib/appsignal/logger.rb
```diff
@@ -126,6 +126,19 @@ def fatal(message = nil, attributes = {})
...
+    # When using ActiveSupport::TaggedLogging without the broadcast feature,
+    # the passed logger is required to respond to the `silence` method.
+    # In our case it behaves as the broadcast feature of the Rails logger, but
+    # we don't have to check if the parent logger has the `silence` method defined
+    # as our logger directly inherits from Ruby base logger.
+    #
+    # Links:
+    # https://github.com/rails/rails/blob/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/activesupport/lib/active_support/logger.rb#L60-L76
+    # https://github.com/rails/rails/blob/main/activesupport/e11ebc04cfbe41c06cdfb70ee5a9fdbbd98bb263/active_support/logger_silence.rb
+    def silence(_severity = ERROR, &block)
```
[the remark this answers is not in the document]
luismiramirez: Yep, I noticed. I'm not sure if we should go back to using `block.call` this method's signature is different from the original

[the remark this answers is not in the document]
tombruijn: Whether you use yield or block doesn't matter for the method signature. It's a Ruby syntax thing only.

PR #1019, opened 2023-12-26
Fix for Ruby 3.3

cmer:
I had to implement this simple fix for my Rails application to run under Ruby 3.3.0.

### DIFF
--- a/lib/appsignal/logger.rb
+++ b/lib/appsignal/logger.rb
@@ -25,6 +25,8 @@ module Appsignal
     def initialize(group, level: INFO, format: PLAINTEXT)
       raise TypeError, "group must be a string" unless group.is_a? String
 
+      super(group, :level => level)
+
       @group = group
       @level = level
       @format = format
--- a/.changesets/fix-compatibility-for-ruby-3-3.md
+++ b/.changesets/fix-compatibility-for-ruby-3-3.md
@@ -1,0 +1,6 @@
+---
+bump: "patch"
+type: "fix"
+---
+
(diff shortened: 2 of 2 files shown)


luismiramirez:
Hey there @cmer!

How's the gem behaving on your app without this fix?

cmer:
Literally nothing works. 500 on everything.

northeastprince:
[Same.](https://github.com/hackclub/hackathons-backend/issues/379)

lib/appsignal/logger.rb
```diff
@@ -23,6 +23,7 @@ class Logger < ::Logger
     # @param level Log level to filter with
     # @return [void]
     def initialize(group, level: INFO, format: PLAINTEXT)
+      super(group, level: level)
```
northeastprince, line 26:
```suggestion
      super(group, level:)
```

cmer: I used the "old syntax" in case they need to support older Ruby versions. I'm not sure how far back they want to support.

northeastprince: Ah, makes sense. You'll need to use hash rocket syntax then to comply with RuboCop.

luismiramirez:
Thanks for the contribution and info @northeastprince and @cmer 

The fix has been published on [version 3.5.1](https://rubygems.org/gems/appsignal/versions/3.5.1)

luismiramirez:
You can find the proper fix on [version 3.5.2](https://rubygems.org/gems/appsignal/versions/3.5.2)

Issue #133, opened 2016-08-31
Make installation automatable


LouisKottmann:
Hello,

Right now when I run `bundle exec appsignal install $APPSIGNAL_KEY` it asks me a bunch of questions that we'd like Ops to handle and not Devs (especially in the case of many apps).

However `bundle exec appsignal install --help` does not offer an option to do this via flags or environment variables, and reading [install.rb](lib/appsignal/cli/install.rb) I don't think it is possible at all for now.

Would you consider adding flags to the `install` action (or reading env vars, whatever works for you)?

Regards

thijsc:
Hi Louis, thanks for trying AppSignal!

Actually one of the questions the installer asks is wether you'd like to use env vars. If you choose that option it will output a list of env vars you can send to your ops person.

This is just for convenience, AppSignal is fully configurable through the environment. You can find the options here: http://docs.appsignal.com/gem-settings/configuration.html

Does that help?

LouisKottmann:
So if I specify these env vars I don't have to run `bundle exec appsignal install`?

thijsc:
That's correct. The installer is just a convenient way to get a config, using it is optional.

LouisKottmann:
Alright thanks!

PR #1037, opened 2024-02-04
Monitor serving of Rack response bodies

julik:
Some work might be getting done within a Rack response body. For example, when ActionController::Streaming is used, or when a Rack app elects to stream a response.

The Rack SPEC doc actually defines the behavior in sufficient detail to wrap this into the same Appsignal transaction.

The Rails instrumentation may also inherit from the more generic one as most methods and constants are shared then

## To do

- [x] Add a changeset before merge

### DIFF
--- a/lib/appsignal/rack/body_wrapper.rb
+++ b/lib/appsignal/rack/body_wrapper.rb
@@ -31,7 +31,7 @@ module Appsignal
           # this is not going to work since the SPEC says that if both are available,
           # `each` should be used and `call` should be ignored.
           # So for that case we can drop by to our default EnumerableBodyWrapper
-          CallableBodyWrapper.new(original_body, appsignal_transaction_or_nil)
+          CallableBodyWrapper.new(original_body, appsignal_transaction)
         else
           EnumerableBodyWrapper.new(original_body, appsignal_transaction)
         end
--- a/lib/appsignal/rack/body_wrapper.rb
+++ b/lib/appsignal/rack/body_wrapper.rb
@@ -42,7 +42,9 @@ module Appsignal
         # The @body_already_closed check is needed so that if `to_ary`
         # of the body has already closed itself (as prescribed) we do not
         # attempt to close it twice
-        @body.close if !@body_already_closed && @body.respond_to?(:close)
+        if !@body_already_closed && @body.respond_to?(:close)
+          Appsignal.instrument("process_action.response_body_close") { @body.close }
+        end
         @body_already_closed = true
       rescue Exception => error # rubocop:disable Lint/RescueException
         @transaction&.set_error(error)
@@ -74,7 +76,7 @@ module Appsignal
         # in a blockless way it is still a good idea to have it in place.
         return enum_for(:each) unless block_given?
 
-        @body.each(&blk)
+        Appsignal.instrument("process_action.response_body_each") { @body.each(&blk) }
       rescue Exception => error # rubocop:disable Lint/RescueException
         @transaction&.set_error(error)
         raise error
@@ -91,7 +93,7 @@ module Appsignal
       def call(stream)
         # `stream` will be closed by the app we are calling, no need for us
         # to close it ourselves
-        @body.call(stream)
+        Appsignal.instrument("process_action.response_body_call") { @body.call(stream) }
       rescue Exception => error # rubocop:disable Lint/RescueException
         @transaction&.set_error(error)
         raise error
@@ -110,7 +112,7 @@ module Appsignal
     class ArrayableBodyWrapper < EnumerableBodyWrapper
       def to_ary
         @body_already_closed = true
-        @body.to_ary
+        Appsignal.instrument("process_action.response_body_to_ary") { @body.to_ary }
       rescue Exception => error # rubocop:disable Lint/RescueException
         @transaction&.set_error(error)
         raise error
@@ -131,7 +133,7 @@ module Appsignal
     # pass that file to the downstream webserver for sending using X-Sendfile
     class PathableBodyWrapper < EnumerableBodyWrapper
       def to_path
-        @body.to_path
+        Appsignal.instrument("process_action.response_body_to_path") { @body.to_path }
       rescue Exception => error # rubocop:disable Lint/RescueException
         @transaction&.set_error(error)
         raise error
--- a/lib/appsignal/rack/body_wrapper.rb
+++ b/lib/appsignal/rack/body_wrapper.rb
@@ -4,7 +4,7 @@ module Appsignal
   # @api private
   module Rack
     class BodyWrapper
-      def self.wrap(original_body, appsignal_transaction_or_nil)
+      def self.wrap(original_body, appsignal_transaction)
         # The logic of how Rack treats a response body differs based on which methods
         # the body responds to. This means that to support the Rack 3.x spec in full
         # we need to return a wrapper which matches the API of the wrapped body as closely
@@ -21,13 +21,13 @@ module Appsignal
         # This comment https://github.com/rails/rails/pull/49627#issuecomment-1769802573
         # is of particular interest to understand why this has to be somewhat complicated.
         if original_body.respond_to?(:to_path)
-          PathableBodyWrapper.new(original_body, appsignal_transaction_or_nil)
+          PathableBodyWrapper.new(original_body, appsignal_transaction)
         elsif original_body.respond_to?(:to_ary)
-          ArrayableBodyWrapper.new(original_body, appsignal_transaction_or_nil)
+          ArrayableBodyWrapper.new(original_body, appsignal_transaction)
         elsif !original_body.respond_to?(:each) && original_body.respond_to?(:call)
-          CallableBodyWrapper.new(original_body, appsignal_transaction_or_nil)
+          CallableBodyWrapper.new(original_body, appsignal_transaction)
         else
-          EnumerableBodyWrapper.new(original_body, appsignal_transaction_or_nil)
+          EnumerableBodyWrapper.new(original_body, appsignal_transaction)
         end
       end
 
@@ -47,15 +47,19 @@ module Appsignal
         end
         @body_already_closed = true
       rescue Exception => error # rubocop:disable Lint/RescueException
-        @transaction&.set_error(error)
+        @transaction.set_error(error)
         raise error
       ensure
+        complete_transaction!
+      end
+
+      def complete_transaction!
         # We need to call the Transaction class method and not
         # @transaction.complete because the transaction is still
         # thread-local and it needs to remove itself from the
         # thread variables correctly, which does not happen on
         # Transaction#complete.
-        Appsignal::Transaction.complete_current! if @transaction
+        Appsignal::Transaction.complete_current! unless @transaction.nil_transaction?
       end
     end
 
@@ -78,7 +82,7 @@ module Appsignal
 
         Appsignal.instrument("response_body_each.rack") { @body.each(&blk) }
       rescue Exception => error # rubocop:disable Lint/RescueException
-        @transaction&.set_error(error)
+        @transaction.set_error(error)
         raise error
       end
     end
@@ -95,7 +99,7 @@ module Appsignal
         # to close it ourselves
         Appsignal.instrument("response_body_call.rack") { @body.call(stream) }
       rescue Exception => error # rubocop:disable Lint/RescueException
-        @transaction&.set_error(error)
+        @transaction.set_error(error)
         raise error
       end
     end
@@ -114,18 +118,12 @@ module Appsignal
         @body_already_closed = true
         Appsignal.instrument("response_body_to_ary.rack") { @body.to_ary }
       rescue Exception => error # rubocop:disable Lint/RescueException
-        @transaction&.set_error(error)
+        @transaction.set_error(error)
         raise error
       ensure
         # We do not call "close" on ourselves as the only action
         # we need to complete is completing the transaction.
-        #
-        # We need to call the Transaction class method and not
-        # @transaction.complete because the transaction is still
-        # thread-local and it needs to remove itself from the
-        # thread variables correctly, which does not happen on
-        # Transaction#complete.
-        Appsignal::Transaction.complete_current! if @transaction
+        complete_transaction!
       end
     end
 
@@ -135,7 +133,7 @@ module Appsignal
       def to_path
         Appsignal.instrument("response_body_to_path.rack") { @body.to_path }
       rescue Exception => error # rubocop:disable Lint/RescueException
-        @transaction&.set_error(error)
+        @transaction.set_error(error)
         raise error
       end
     end
--- a/lib/appsignal/rack/body_wrapper.rb
+++ b/lib/appsignal/rack/body_wrapper.rb
@@ -1,0 +1,128 @@
+# frozen_string_literal: true
+
+module Appsignal
+  # @api private
+  module Rack
+    class BodyWrapper
+      def self.wrap(original_body, appsignal_transaction_or_nil)
+        # The logic of how Rack treats a response body differs based on which methods
+        # the body responds to. This means that to support the Rack 3.x spec in full
+        # we need to return a wrapper which matches the API of the wrapped body as closely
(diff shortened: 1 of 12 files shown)


julik:
@tombruijn @jeffkreeftmeijer this is a draft - wanted to ask you what you think. I have encountered in a few cases that this would be very useful, as currently streaming (or non-presized) bodies elide the instrumentation _and_ (more importantly) do not send errors.

julik:
This will invariably move things closer for https://github.com/appsignal/appsignal-ruby/issues/329 too

tombruijn:
Hi @julik, this change is looking good so far 👍 
It would be helpful if you could split the refactor to the generic middleware and the changes to support streaming bodies into separate commits.
A small (rails?) test app that highlights the problem of it not reporting errors would also be helpful for us to review it. We have [a repo with test apps](https://github.com/appsignal/test-setups). You can modify one of the example apps, but a gist or separate repo would also be fine.
We'll handle the deprecation of constants.

julik:
Ok, let me get the PR to the place where all tests pass and all the bodies are tested, then we can take a look how to split it into two commits in a more logical way.

julik:
Limited to just adding the body, LMK if you want the class hierarchy change in a different PR

julik:
I am curious why `transaction.set_http_or_background_queue_start` gets called after calling into the app though, but I don't understand the mechanics well enough.

tombruijn:
> I am curious why `transaction.set_http_or_background_queue_start` gets called after calling into the app though, but I don't understand the mechanics well enough.

It doesn't really matter when it's called. The start time we read comes from a header, which doesn't change during the request. We mostly call it at the end when all the other metadata is also set.

julik:
Okido. This should be good to go now.

tombruijn:
Thanks for the updates!

Do you have an app or some part of a controller that we can test your changes with?

So far, testing it on our [ruby/rails7-sidekiq app](https://github.com/appsignal/test-setups/tree/main/ruby/rails7-sidekiq/app), I'm getting no data from Rails requests.

Getting a lot of these logs from the [transaction initializer](https://github.com/appsignal/appsignal-ruby/blob/4c2845cedf6399b973759e31a539a415d4bb565b/lib/appsignal/transaction.rb#L29-L34), so it's a bit confused at the moment about which transaction is active. This happens when a transaction is being created when another is active, or viewed in another way, when the previous request's transaction hasn't been completed.

> [2024-02-06T13:29:34 (process) #133][DEBUG] Trying to start new transaction with id '01aef6e8-2576-4dce-9d79-777263c44de9', but a transaction with id '78b895a4-778a-41c0-855d-1866b8bf1a5b' is already running. Using transaction '78b895a4-778a-41c0-855d-1866b8bf1a5b'.

lib/appsignal/rack/body_wrapper.rb
```diff
@@ -0,0 +1,128 @@
...
+        # The @body_already_closed check is needed so that if `to_ary`
+        # of the body has already closed itself (as prescribed) we do not
+        # attempt to close it twice
+        @body.close if !@body_already_closed && @body.respond_to?(:close)
+        @body_already_closed = true
+      rescue Exception => error # rubocop:disable Lint/RescueException
+        @transaction&.set_error(error)
+        raise error
+      ensure
+        @transaction&.complete
```
tombruijn, line 51: You need to call `Appsignal::Transaction.complete_current!` so it unsets the current transaction stored in the Thread variable.

https://github.com/appsignal/appsignal-ruby/blob/4c2845cedf6399b973759e31a539a415d4bb565b/lib/appsignal/transaction.rb#L65-L74

julik:
> So far, testing it on our [ruby/rails7-sidekiq app](https://github.com/appsignal/test-setups/tree/main/ruby/rails7-sidekiq/app), I'm getting no data from Rails requests.

I will take a look what it could be. I checked out the repo and I was wondering what API key I should be using if I want to run one of the apps there, and how do the actual requests to the apps get done? does it use Net::HTTP or I am supposed to visit the pages by myself?

> Getting a lot of these logs from the [transaction initializer](https://github.com/appsignal/appsignal-ruby/blob/4c2845cedf6399b973759e31a539a415d4bb565b/lib/appsignal/transaction.rb#L29-L34), so it's a bit confused at the moment about which transaction is active. This happens when a transaction is being created when another is active, or viewed in another way, when the previous request's transaction hasn't been completed.

That's probably the complete bit you mentioned earlier. But why would `Transaction#complete` here not work?

julik:
> Do you have an app or some part of a controller that we can test your changes with? That would help ensure we're reviewing the things your fixing/changing.

I might actually try it with Cheddar briefly 😆  but let me see whether I can get the `complete_current` to work

unflxw:
> what API key I should be using if I want to run one of the apps there

Your organization-level API key should be fine. 

> how do the actual requests to the apps get done?

You go to http://localhost:4001 and click around. These are primarily for manual testing.

julik:
> These are primarily for manual testing

Aight. I'll click around then.

tombruijn:
> > Do you have an app or some part of a controller that we can test your changes with? That would help ensure we're reviewing the things your fixing/changing.
> 
> I might actually try it with Cheddar briefly 😆 but let me see whether I can get the `complete_current` to work

When I changed [the line 51 I commented on in this comment](https://github.com/appsignal/appsignal-ruby/pull/1037#discussion_r1479819517) with `Appsignal::Transaction.complete_current!` it started reporting data.

I hope you see the same behavior in your app. Check the `appsignal.log` file's debug messages. You will need to set the [`log_level` config option](https://docs.appsignal.com/ruby/configuration/options.html#option-log_level) to `debug` to see those logs.

I'd also to test a streaming request hitting this new code.

julik:
@ioquatix apologies for summoning, but could you help me check this? We want the `BodyWrapper` to be compatible across Rack 1, 2 and 3 (and beyond, if at all possible). I followed the tribulations in the rack and rails repo with interest and tried to make the solution ultra-safe here. Is there anything that cold crash here? esp. the semantics of when to call / not to call `body.close`

ioquatix:
This is a tricky thing to implement.

https://github.com/rack/rack/blob/main/lib/rack/body_proxy.rb and similar for older versions of Rack would be the gold standard for "proxy a body".

The expectation is you'd define `#close` and call super if possible, and implement your own logic there too.

Long running responses could run for minutes or hours or longer. Is that okay?

If you want validation, you should test on Rack 1, 2 and 3, using `Rack::Lint`.

julik:
Thanks for your input @ioquatix it does seem that with the fact that we need to `rescue` to capture errors correctly the implementation that I came up with is not as overkill as it seemed.

julik:
@tombruijn this is ready to go as far as I am concerned. Does lintje have an autofix option? I've added test code to https://github.com/appsignal/test-setups/pull/186 but I haven't tested them all, I presume it's all working fine to be honest. If you could give the test setups a second look I would be grateful.

julik:
Also: one could in theory wrap the serving out of the body (what's inside `each`) into an `instrument`, but that's trivial to do at this stage - leaving that decision to you.

julik:
FWIW we've deployed this and looks like it works:

Screen Shot 2024-02-13 at 22 42 08

Screen Shot 2024-02-13 at 22 42 56

This is fairly typical exception occurring in streaming responses. The rest of the application is working perfectly.

julik:
Well those were the famous last words. _Now_ it should be good to go. I can squash the commits before merging.

tombruijn:
Testing with the test apps in PR https://github.com/appsignal/test-setups/pull/187, show the following timeline, which means it's working:

image

Let the review being!

lib/appsignal/rack/generic_instrumentation.rb
```diff
@@ -16,7 +16,8 @@ def call(env)
         if Appsignal.active?
           call_with_appsignal_monitoring(env)
         else
-          @app.call(env)
+          status, headers, obody = @app.call(env)
+          [status, headers, Appsignal::Rack::BodyWrapper.wrap(obody, _transaction = nil)]
```
tombruijn, line 20: A pattern we use is using [`Appsignal::Transaction::NilTransaction`](https://rubydoc.info/gems/appsignal/3.5.6/Appsignal/Transaction/NilTransaction) when no transaction is active when calling `Appsignal::Transaction.current`. Appsignal::Transaction::NilTransaction is * not actually a nil object *, but responds to any method call and ignores it. Which is great for when you don't want your implementation in BodyWrapper to constantly check if the transaction variable is not nil.

Would that work here too, and in the other middlewares where the same thing happens? Pass in `Appsignal::Transaction::NilTransaction.new` and remove all the nil-checks from the BodyWrapper implementation.

lib/appsignal/rack/generic_instrumentation.rb
```diff
@@ -27,19 +28,30 @@ def call_with_appsignal_monitoring(env)
...
+            status, headers, obody = @app.call(env)
+            [status, headers, Appsignal::Rack::BodyWrapper.wrap(obody, transaction)]
           end
         rescue Exception => error # rubocop:disable Lint/RescueException
           transaction.set_error(error)
+          complete_transaction_without_body = true
           raise error
         ensure
-          transaction.set_action_if_nil(env["appsignal.route"] || "unknown")
+          default_action = env["appsignal.route"] || env["appsignal.action"] || "unknown"
```
tombruijn, line 46: We should standardize this to only `appsignal.action` 😬 
We'll do that another time, not in this PR.

lib/appsignal/rack/generic_instrumentation.rb
```diff
@@ -27,19 +28,30 @@ def call_with_appsignal_monitoring(env)
           Appsignal::Transaction::HTTP_REQUEST,
           request
         )
+        # We need to complete the transaction if there is an exception exception inside the `call`
```
tombruijn, line 31:
```suggestion
        # We need to complete the transaction if there is an exception inside the `call`
```

tombruijn (commented):
Looking really good so far! Here's a couple thoughts. Don't feel the need to immediately make changes. Let's first discuss.

lib/appsignal/rack/generic_instrumentation.rb
julik: We could use the `NilTransaction`, sure. What threw me off there was that there would be checks anyway because I don't know what will happen when you call `complete_current!` (or rather: it is possible to find out what happens, but I don't know whether that behaviour is stable) - so `transaction.complete` cannot be used, which would have been nice. If we can remove checks entirely and `complete_current!` will do the thing gracefully - even better.

lib/appsignal/rack/generic_instrumentation.rb
julik: Okay, I will disregard 😆

lib/appsignal/rack/body_wrapper.rb
```diff
@@ -0,0 +1,143 @@
...
+        @transaction = appsignal_transaction
+      end
+
+      # This must be present in all Rack bodies and will be called by the serving adapter
+      def close
+        # The @body_already_closed check is needed so that if `to_ary`
+        # of the body has already closed itself (as prescribed) we do not
+        # attempt to close it twice
+        if !@body_already_closed && @body.respond_to?(:close)
+          Appsignal.instrument("process_action.response_body_close") { @body.close }
```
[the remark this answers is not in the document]
julik: And for the others? `"response_body_each.rack"`, `"response_body_to_a.rack"` and so on?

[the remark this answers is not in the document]
tombruijn: Yes, all `Appsignal.instrument` calls need to be updated.

tombruijn (approved):

lib/appsignal/rack/body_wrapper.rb
```diff
@@ -0,0 +1,141 @@
...
+        complete_transaction!
+      end
+
+      def complete_transaction!
+        # We need to call the Transaction class method and not
+        # @transaction.complete because the transaction is still
+        # thread-local and it needs to remove itself from the
+        # thread variables correctly, which does not happen on
+        # Transaction#complete.
+        Appsignal::Transaction.complete_current! unless @transaction.nil_transaction?
```
tombruijn, line 62: It shouldn't matter if you call `Appsignal::Transaction.complete_current!` when it's a `NilTransaction`.

I think you can leave out the condition.

```suggestion
        Appsignal::Transaction.complete_current!
```

I don't know if there would be a scenario where the given transaction to this wrapper is empty but there is a "current transaction" set on the Thread.

tombruijn:
Let's wait for one more review before merging.

@julik can you fix the last Rubocop issues?

julik:
> @julik can you fix the last Rubocop issues?

Done!

luismiramirez (approved):

unflxw (approved):
Looks good to me! The test setup works fine. I have some minor nitpicks about how the information is presented.

These are very uninformed suggestions, in that I'm not a Rack expert by any means, nor am I familiar with the streaming use case and which information would be useful in it.

That is, if some of the information that I suggest removing is actually useful, feel free to ignore the suggestion.

---

- Instead of `response_body_each`, I think the category name might work better as a verb, like "process_action" or "query"? I'm thinking `stream_response`?
- I don't know whether there's much value on having different category names for each internal implementation of the body (`response_body_each`, `response_body_call`, ...) I feel that this is more likely to confuse folks than to inform them?
- As the cherry on top, it would be nice if the category name wasn't too long -- very long category names eat into the event timeline in the UI.

**Suggestion:** use a single category for all the response body instrumentation blocks (`stream_response.rack`?) and, if the information about the specific implementation that's being used is relevant, add that as part of the title, which shows up on the UI on hover:

```ruby
  Appsignal.instrument(
    "stream_response.rack",
    "Stream response body (#to_path)"
  ) { @body.to_path }
```

--- 

- Regarding the body close event: is this a relevant event to instrument? I believe that this event would _always_ happen right after the body streaming event. Does this event add any information to the event timeline that isn't already added by the finishing time of the body streaming event?

**Suggestion**: do not instrument the call to `.close` on the body.

julik:
> Does this event add any information to the event timeline that isn't already added by the finishing time of the body streaming event?

I'd vouch it does, because there can be operations there which take a long time. For example - deleting temp data, checking in connections into a connection pool, running some kind of `after_commit` activity etc. When Rack middleware gets developed and one needs to reach for the `close` method that method often does some kind of "flushing" or "cleanup" which might be expensive. The absence of a span for it in the waterfall could also indicate that `close` didn't get called, which would be a bug in the webserver.

> use a single category for all the response body instrumentation blocks

I'll leave it to @tombruijn be the judge on that, no objections from me. I separated them so that it would be apparent which response body method got called - this is related to the Rails bugs linked inline from the comments, but no strong opinion.

ioquatix:
Instrumenting `response#close` is probably valuable if you capture it as a span.

tombruijn:
There's a lot of technical stuff in event names now already from libraries like Rails that's not always obvious, but we can make the experience a little easier now that we have the option. The event name should be something `process_response_body.rack` then. We we can add the details about what type in the event title, like "Process Rack response body (#to_path)", unless it's crucial information you need at a glance looking at the timeline.

And let's keep the close event.

@julik can you update the event categories and add the event titles if you agree?

tombruijn:
Event names:
- Close event:
    - Category: `response_body_close.rack`
    - Title: None
- EnumerableBodyWrapper event:
    - Category: `process_response_body.rack`
    - Title: `Process Rack response body (#each)`
- CallableBodyWrapper event:
    - Category: `process_response_body.rack`
    - Title: `Stream response body (#call)`
- ArrayableBodyWrapper event:
    - Category: `process_response_body.rack`
    - Title: `Stream response body (#to_ary)`
- PathableBodyWrapper event:
    - Category: `response_body_to_path.rack`
    - Title: None

I asked Julik a question on Slack about the `to_path` event naming, otherwise I think should be good.

tombruijn:
Thanks for the PR @julik 🎉 This will be included in the next release, which will be a minor version bump. I think we can publish it tomorrow.

tombruijn:
This has been released in Ruby gem 3.6.0 🚀

PR #1083, opened 2024-05-14
Do not use `trace` on internal logger

unflxw:
The internal logger does not support `trace`. Use `debug` instead.

Since the heartbeat implementation intentionally swallows all errors and logs them, issues such as this would go unnoticed by the test suite. Add assertions on the messages that are expected to be logged to ensure that the implementation is actually working as expected.

(Another "how did this ever pass the tests?" mystery, though in this case the answer is simple in retrospect)


unflxw:
Will merge without reviews as to release a fix to the affected customer.

PR #1098, opened 2024-06-18
Skip RubyGems and Bundler update on JRuby

unflxw:
When running on JRuby, skip updating RubyGems and Bundler. This is a workaround for a build issue involving JRuby: https://github.com/jruby/jruby/issues/7059

[skip changeset] because it only affects CI

### DIFF
--- a/support/install_deps
+++ b/support/install_deps
@@ -4,6 +4,12 @@ set -eu
 
 gem_args="--no-verbose --no-document"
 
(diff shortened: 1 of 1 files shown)


tombruijn (approved):

unflxw:
I have manually verified that replacing `"rails"` with `"appsignal"` in the [minimal reproduction case](https://github.com/jruby/jruby/issues/7059#issuecomment-2176005415) does not trigger this issue.

PR #1159, opened 2024-07-05
Fix all_specs deprecation of bundler

jbockler:
bundler deprecated the usage of `Bundler.rubygems.all_specs` with version 2.5.12 (June 13, 2024), in order to better support default gems.
This version is the new default version with ruby 3.3.3 and thus the warning pops up:

Bundler.rubygems.all_specs has been removed in favor of Bundler.rubygems.installed_specs


tombruijn (approved):
Great! Thanks for the PR!

PR #1243, opened 2024-08-14
Remove heartbeat aliases and warnings

unflxw:
Remove all name and argument aliases for the deprecated heartbeat invocations, and the corresponding deprecation warnings.

[skip changeset]


tombruijn (approved):
Thanks!

PR #1293, opened 2024-09-12
Implement heartbeat checkins

unflxw:
Part of https://github.com/appsignal/integration-guide/issues/169.

### [Ensure stubs are called exactly once](https://github.com/appsignal/appsignal-ruby/commit/1971bbe2fae6566c1b1df0ea88bfd95c926c608f)

_[@tombruijn could you take a look? I tried more reasonable approaches within webmock before reaching this count-requests-separately solution, but maybe I missed something in webmock that would solve this in a less awkward way]_

Ensure that the stubs are called exactly once -- that is, that the
same stub is not used several times, potentially masking a bug where
several transmissions happen for the same events, and that no stubs
are declared which are left unused.

The strange way in which it is implemented is due to certain quirks
of how the `webmock` gem works. I could not find a way to have a
stub automatically disable itself when called, similar to what the
`nock` Node.js package provides -- so I implemented this by passing
a number of request bodies to expect to respond to when declaring
the mock, and erroring if more requests are made than the request
bodies provided.

That, however, does not cover the scenario where the stub is not
called at all. Somewhat bafflingly, the `have_been_requested` matcher
does not check a counter that has been incremented for each request.
Instead, it re-runs the requests it has stored by the stubs _again_,
and counts how many match. Given the implementation described above,
where the stub is exhausting a list of request bodies as it goes,
the matcher will fail when attempting to go a second time around.

So instead, we increment a counter when a request is processed
(without said request having exhausted the matcher) and we return
that counter from the stub. The tests store these counters in a
`let` binding, and an `after` block checks that each stored counter
has been increased to one. This is done in an `after` block not
only for convenience, but because some tests are mocking a call
that will only be performed in that same `after` block, when the
scheduler is stopped.

### [Implement heartbeat check-ins](https://github.com/appsignal/appsignal-ruby/commit/2abbd5e2f60bcb11b30d7ae2c68ac58dd35377fa)

Add an `Appsignal::CheckIn.heartbeat` helper that emits a single
heartbeat for the check-in identifier given.

When given `continuous: true` as the second argument, it spawns a
separate thread that emits a heartbeat every thirty seconds. This
is a convenience method for the use case where the heartbeat is
only meant as a check that the process is alive.

Split functions that deal with different event kinds out of the
scheduler and test them independently.

### DIFF
--- a/lib/appsignal/check_in.rb
+++ b/lib/appsignal/check_in.rb
@@ -2,10 +2,21 @@
 
 module Appsignal
   module CheckIn
+    HEARTBEAT_CONTINUOUS_INTERVAL_SECONDS = 30
     class << self
+      # @api private
+      def continuous_heartbeats
+        @continuous_heartbeats ||= []
+      end
+
+      # @api private
+      def kill_continuous_heartbeats
+        continuous_heartbeats.each(&:kill)
+      end
+
       # Track cron check-ins.
       #
-      # Track the execution of certain processes by sending a cron check-in.
+      # Track the execution of scheduled processes by sending a cron check-in.
       #
       # To track the duration of a piece of code, pass a block to {.cron}
       # to report both when the process starts, and when it finishes.
@@ -40,6 +51,37 @@ module Appsignal
         output
       end
 
+      # Track heartbeat check-ins.
+      #
+      # Track the execution of long-lived processes by sending a heartbeat
+      # check-in.
+      #
+      # @example Send a heartbeat check-in
+      #   Appsignal::CheckIn.heartbeat("main_loop")
+      #
+      # @param identifier [String] identifier of the heartbeat check-in to report.
+      # @param continuous [Boolean] whether the heartbeats should be sent continuously
+      #   during the lifetime of the process. Defaults to `false`.
+      # @yield the block to monitor.
+      # @return [void]
+      # @since 4.1.0
+      # @see https://docs.appsignal.com/check-ins/heartbeat
+      def heartbeat(identifier, continuous: false)
+        if continuous
+          continuous_heartbeats << Thread.new do
+            loop do
+              heartbeat(identifier)
+              sleep HEARTBEAT_CONTINUOUS_INTERVAL_SECONDS
+            end
+          end
+
+          return
+        end
+
+        event = Event.heartbeat(:identifier => identifier)
+        scheduler.schedule(event)
+      end
+
       # @api private
       def transmitter
         @transmitter ||= Transmitter.new(
@@ -60,5 +102,6 @@ module Appsignal
   end
 end
 
+require "appsignal/check_in/event"
 require "appsignal/check_in/scheduler"
 require "appsignal/check_in/cron"
--- a/spec/support/helpers/api_request_helper.rb
+++ b/spec/support/helpers/api_request_helper.rb
@@ -1,4 +1,16 @@
 module ApiRequestHelper
+  class RequestCounter
+    attr_reader :count
+
+    def initialize
+      @count = 0
+    end
+
+    def increase
+      @count += 1
+    end
+  end
+
   def stub_api_request(config, path, body = nil)
     options = {
       :query => {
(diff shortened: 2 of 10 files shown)


lib/appsignal/check_in.rb
```diff
@@ -2,10 +2,18 @@
...
 module Appsignal
   module CheckIn
+    HEARTBEAT_FOREVER_INTERVAL_SECONDS = 30
     class << self
+      @@forever_heartbeats = [] # rubocop:disable Style/ClassVars
+
+      # @api private
+      def kill_forever_heartbeats
+        @@forever_heartbeats.each(&:kill)
+      end
```
tombruijn, line 12: I'm not sure why the class env var is needed here.

This will work the way you want it I think, a single list of these forever heartbeats kept on the module.

```suggestion
      # @api private
      def forever_heartbeats
        @forever_heartbeats ||= []
      end

      # @api private
      def kill_forever_heartbeats
        forever_heartbeats.each(&:kill)
      end
```

lib/appsignal/check_in.rb
```diff
@@ -40,6 +48,35 @@ def cron(identifier)
...
+      #   Appsignal::CheckIn.heartbeat("main_loop")
+      #
+      # @param identifier [String] identifier of the heartbeat check-in to report.
+      # @yield the block to monitor.
+      # @return [void]
+      # @since 4.1.0
+      # @see https://docs.appsignal.com/check-ins/heartbeat
+      def heartbeat(identifier, forever: false)
+        if forever
+          @@forever_heartbeats << Thread.new do
```
tombruijn, line 66:
```suggestion
          forever_heartbeats << Thread.new do
```

spec/support/helpers/api_request_helper.rb
```diff
@@ -1,4 +1,16 @@
...
+    attr_reader :count
+
+    def initialize
+      @count = 0
+    end
+
+    def increase
+      @count += 1
+    end
+  end
```
tombruijn, line 12: > @tombruijn could you take a look? I tried more reasonable approaches within webmock before reaching this count-requests-separately solution, but maybe I missed something in webmock that would solve this in a less awkward way

I'll take a look later. Your implementation seems to work, but a bit weird.

lib/appsignal/check_in.rb
```diff
@@ -40,6 +48,35 @@ def cron(identifier)
...
+      #
+      # @example Send a heartbeat check-in
+      #   Appsignal::CheckIn.heartbeat("main_loop")
+      #
+      # @param identifier [String] identifier of the heartbeat check-in to report.
+      # @yield the block to monitor.
+      # @return [void]
+      # @since 4.1.0
+      # @see https://docs.appsignal.com/check-ins/heartbeat
+      def heartbeat(identifier, forever: false)
```
tombruijn, line 64: Was the name forever decided upon yet? We can also name it continuous heartbeats or something. Forever seems so definitive.

tombruijn (approved):

lib/appsignal/check_in.rb
unflxw: I think I tried this and it didn't work -- I'll double-check. 👍

lib/appsignal/check_in.rb
tombruijn: > It's not at all set in stone. Something like `continously: true` sounds good to me as well.

My vote is for something other than forever, like `repeat: true` or `continuous: true`.

lib/appsignal/check_in.rb
tombruijn: I tried it locally before commenting and it worked for me :)

unflxw: Ah, indeed -- what didn't work was trying to initialise it as `@forever_heartbeats = []` within `class << self` itself.

lib/appsignal/check_in.rb
unflxw: Went with `continuous: true`.

lib/appsignal/check_in.rb
```diff
@@ -40,6 +51,37 @@ def cron(identifier)
...
+      #
+      # Track the execution of long-lived processes by sending a heartbeat
+      # check-in.
+      #
+      # @example Send a heartbeat check-in
+      #   Appsignal::CheckIn.heartbeat("main_loop")
+      #
+      # @param identifier [String] identifier of the heartbeat check-in to report.
+      # @param continuous [Boolean] whether the heartbeats should be sent continuously
+      # during the lifetime of the process. Defaults to `false`.
```
tombruijn, line 64:
```suggestion
      #   during the lifetime of the process. Defaults to `false`.
```

Needs to be indented otherwise it becomes part of the description of the entire method:


2024-09-16 10-11-53 Module AppsignalCheckIn — Documentation for ruby(1)

PR #1334, opened 2024-11-12
Add default_attributes to Logger

jvanbaarsen:
This adds the ability to add default attributes to a logger object.

One usecase for this:

You create a Logger, for which you already know a specific attribute, you can now set it on the initialize instead of having to pass it around all the time.


unflxw (approved):
Pushed some additional changes -- most notably I renamed `default_attributes` to `attributes`.

Issue #1352, opened 2024-12-18
Tagged logging support breaks Rails logger broadcasting


unflxw:
We implemented support for ActiveRecord tagged logging in our logger #1344.

Due to bugs in Rails' broadcast implementation, broadcasting to several tagged loggers will break tagged logging functionality (https://github.com/rails/rails/issues/49745) and cause Rails itself to return malformed Rack requests (https://github.com/rails/rails/issues/51883)

There is a PR in Rails for fixing this by having the broadcast logger explicitly support tagged logging (https://github.com/rails/rails/pull/53105) but it doesn't seem like it'll be merged soon.

To avoid breaking broadcast logging, we may need to implement our own workaround to this issue.

PR #1357, opened 2024-12-19
Implement logger broadcasting

unflxw:
This PR builds on top of the changes in #1355.

Docs will need to be updated to recommend this over `Rails.logger.broadcast_to`.

---

### [Implement logger broadcasting](https://github.com/appsignal/appsignal-ruby/commit/8d31106218c56df1c7a7044ffcc73b72da586821)

Implement broadcasting logs to other loggers, following the same
interface as `#broadcast_to`. Unlike `#broadcast_to`, however,
instead of forwarding each method call one-for-one, we rely on the
`#add` method which is expected of the logger interface.

As such, we should not run into the kinds of bugs that affect
`ActiveSupport::BroadcastLogger` when used alongside additional
logger features such as tagged logging, such as https://github.com/rails/rails/issues/49745.

This enables customers to use both broadcast logging and tagged
logging with the AppSignal logger without running into
application-breaking bugs.

As a trade-off, unlike `ActiveSupport::BroadcastLogger`, our
broadcast implementation is sensitive to the order in which those
additional features are applied, expecting them to be applied on
top of the broadcast logger, and not on the underlying loggers:

```ruby
 # Incorrect example
some_logger = ActiveSupport::TaggedLogging.new(
  Logger.new("/etc/app.log")
)

appsignal_logger = Appsignal::Logger.new("app")

 # Even if `some_logger` is a tagging-enabled logger,
 # this logger will not behave as a tagging-enabled logger.
Rails.logger = appsignal_logger.broadcast_to(some_logger)
```

```ruby
 # Correct example
some_logger = Logger.new("/etc/app.log")
appsignal_logger = Appsignal::Logger.new("app")

appsignal_logger.broadcast_to(some_logger)
 # As the tagged logging is applied on the broadcast logger,
 # both logs sent to AppSignal and to `some_logger` will include
 # given tags, and `Rails.logger` will be a tagging-enabled logger.
Rails.logger = ActiveSupport::TaggedLogging.new(appsignal_logger)
```


luismiramirez (approved):

lightyrs:
@unflxw Doesn't this make it impossible to use (for example) WARN in your appsignal logs and INFO in your file logs?

Here's how we wanted to log:

```rb
  file_logger = Logger.new(Rails.root.join("log/#{Rails.env}.log"), level: Logger::INFO)
  appsignal_logger = Appsignal::Logger.new("rails", level: Logger::WARN)

  config.logger = ActiveSupport::BroadcastLogger.new(
    ActiveSupport::TaggedLogging.new(file_logger),
    ActiveSupport::TaggedLogging.new(appsignal_logger)
  )
```

but then we hit the issues referenced in https://github.com/appsignal/appsignal-ruby/issues/1352

so we switched to:

```rb
  file_logger = Logger.new(Rails.root.join("log/#{Rails.env}.log"), level: Logger::WARN)
  appsignal_logger = Appsignal::Logger.new("rails", level: Logger::WARN)

  appsignal_logger.broadcast_to(file_logger)

  config.logger = ActiveSupport::TaggedLogging.new(appsignal_logger)
```

but ideally we could use INFO for file_logger and WARN for appsignal_logger, like:

```rb
  file_logger = Logger.new(Rails.root.join("log/#{Rails.env}.log"), level: Logger::INFO)
  appsignal_logger = Appsignal::Logger.new("rails", level: Logger::WARN)

  appsignal_logger.broadcast_to(file_logger)

  config.logger = ActiveSupport::TaggedLogging.new(appsignal_logger)
```

PR #1369, opened 2025-01-21
Fix HTTP.rb non-ascii URI parsing

stephannv:
When Appsignal is enabled, it hooks into HTTP request method. It uses `URI.parse` when uri is a string, but `URI.parse` doesn't support non-ascii characters:
```ruby
HTTP.get("http://www.example.com/áéíóúãÔù") # it works when appsignal is disabled
```

With Appsignal enabled, the code above raises this error:
```
URI::InvalidURIError:
       URI must be ascii only "http://www.example.com/\u00E1\u00E9\u00ED\u00F3\u00FA\u00E3\u00D4\u00F9"
```

The solution I found was using Addressable::URI.parse since it handles non-ascii characters.

ps.: Addressable is a HTTP's runtime dependency since 2015

### DIFF
--- a/lib/appsignal/integrations/http.rb
+++ b/lib/appsignal/integrations/http.rb
@@ -5,7 +5,7 @@ module Appsignal
     # @api private
     module HttpIntegration
       def request(verb, uri, opts = {})
-        parsed_request_uri = uri.is_a?(URI) ? uri : URI.parse(uri.to_s)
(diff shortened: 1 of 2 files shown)


unflxw:
Thanks for fixing this issue @stephannv! I've made some minor tweaks to it in #1370 -- continuing there.

PR #1370, opened 2025-01-22
Fix HTTP.rb when using non-ascii URIs

unflxw:
Follow up to #1369 (thanks @stephannv!)

### [Fix HTTP.rb non-ascii URI](https://github.com/appsignal/appsignal-ruby/commit/3eba6738541380d536073f11c8746837588cb523)

When Appsignal is enabled, it hooks into HTTP request method. It
uses `URI.parse` when uri is a string, but `URI.parse` doesn't
support non-ascii characters:

```ruby
HTTP.get("http://www.example.com/áéíóúãÔù")
```

With Appsignal enabled, the code above raises this error:

```
URI::InvalidURIError:
       URI must be ascii only
```

The solution I found was using Addressable::URI.parse since it
handles non-ascii characters.

ps.: Addressable is a HTTP's runtime dependency since 2015

### [Use HTTP::URI if defined](https://github.com/appsignal/appsignal-ruby/commit/b1e64578588c0daa62d44ddcb6d1f96e8d1b19c8)

The `http` gem exposes the `HTTP::URI` module, which on different
versions of the gem either subclasses or includes `Addressable::URI`.

Use it instead of referring directly to `Addressable::URI`, avoiding
explicitly depending on the implementation detail that `HTTP::URI`
uses `Addressable::URI` internally.

When `HTTP::URI` is not defined (in `http` gem versions before 0.8.0)
fall back to using `URI` instead.

### DIFF
--- a/.changesets/fix-http-rb-integration-for-non--ascii-urls.md
+++ b/.changesets/fix-http-rb-integration-for-non--ascii-urls.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+Fix an issue where the HTTP.rb gem integration would raise an error when a string containing non-ASCII characters is passed to the gem as the URL. 
--- a/lib/appsignal/integrations/http.rb
+++ b/lib/appsignal/integrations/http.rb
@@ -5,7 +5,8 @@ module Appsignal
     # @api private
     module HttpIntegration
       def request(verb, uri, opts = {})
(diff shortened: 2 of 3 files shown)


luismiramirez (approved):

tombruijn (approved):
Thanks for the PR and thanks Noemi for the tweaks to it :)
Please also add a changeset

PR #1435, opened 2025-07-24
Change backtrace logging from debug to error level

shairyar:
Updates to error level instead of debug level for better visibility of critical failures.


.changesets/updates-to-error-level-instead-of-debug-level-for-better-visibility-of-critical-failures-.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: change
+---
+
+Use error level for displaying backtraces instead for better visibility of critical failures.
```
unflxw, line 6: Sorry to be annoying, but: error level what? Instead of what?
```suggestion
When an error occurs while initialising AppSignal or when running a probe, use error level logging to log the error's backtrace.
```

Issue #1442, opened 2025-08-14
Incompatible with Async::Cable?


andreaslillebo:
## Describe the bug

We are trying to switch from `puma` to `falcon`. However, we get the following backtrace for requests to `/cable` (ActionCable):
```
Started GET "/cable" for ::1 at 2025-08-14 11:43:55 +0200
 16.1s     warn: Async::Task: Reading HTTP/1.1 requests for Async::HTTP::Protocol::HTTP1::Server. [oid=0x2758] [ec=0x2750] [pid=23607] [2025-08-14 11:43:55 +0200]
               | Task may have ended with unhandled exception.
               |   NoMethodError: undefined method 'each' for an instance of Method (NoMethodError)
               |   
               |           @body.__send__(method_name, *args, &block)
               |                ^^^^^^^^^
               |   → /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/body_proxy.rb:56 in 'Rack::BodyProxy#method_missing'
               |     /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/body_proxy.rb:56 in 'Rack::BodyProxy#method_missing'
               |     /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/body_proxy.rb:56 in 'Rack::BodyProxy#method_missing'
               |     /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/body_proxy.rb:56 in 'Rack::BodyProxy#method_missing'
               |     /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/body_proxy.rb:56 in 'Rack::BodyProxy#method_missing'
               |     /Users/andreaslillebo/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/rack-3.2.0/lib/rack/events.rb:91 in 'Rack::Events::EventedBodyProxy#each'
[...]

PR #1444, opened 2025-08-27
Only instrument Rails if AppSignal started

unflxw:
Issue #1442 can be reproduced without AppSignal starting or being active in the current environment, because the Railtie will add Rails instrumentation even if AppSignal is not active.

This should not happen -- no instrumentation should be added to the customer's application if AppSignal is not active.

Unfortunately, it is not possible to modify the middleware chain at `after_initialize`, so this only fixes the issue for `on_load`.

Note that this does not fix the issue described in #1442.

### DIFF
--- a/.changesets/avoid-instrumenting-rails-when-appsignal-is-not-active.md
+++ b/.changesets/avoid-instrumenting-rails-when-appsignal-is-not-active.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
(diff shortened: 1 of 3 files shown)


unflxw:
This needs to be tested with `start_at` set to `after_initialize`, as it may be too late to modify the middleware stack.

(this is unrelated to the current CI failures)

tombruijn (approved):

PR #1463, opened 2025-09-30
Handle unowned files for `code_ownership`

fatkodima:
Follow up to #1443.

It is perfectly fine to have unowned files, but appsignal currently produces log lines for this case, which look like some errors and can be confusing 

```
INFO -- : ║ [2025-09-30T18:13:32 (process) #8][ERROR] appsignal: Error while looking up CodeOwnership team: NoMethodError: undefined method `name' for nil
```


unflxw (approved):
Looks good to me. Let's make sure to add a changeset for it after merging.

tombruijn (approved):
Thanks!

+1 on adding the changeset after merge.
Let's merge the PR with a squash commit that contains the PR description as the commit message.

PR #1465, opened 2025-10-01
Fix code_ownership for transactions without errors

fatkodima:
Found another case after #1463 was merged.

AppSignal currently produces log lines if transaction has no errors:
```
#125][ERROR] appsignal: Error while looking up CodeOwnership team: NoMethodError: undefined method `backtrace' for nil
```

Follow-up to #1443.


lipskis (approved):
Thanks @fatkodima!

I will merge the PR with a squash commit and fix the line length issue that the git linter is complaining about.

PR #1481, opened 2025-11-20
Add support for ActiveSupport::EventReporter

jeffkreeftmeijer:
Rails 8.1 adds `ActiveSupport::EventReporter`, an interface for reporting structured events. To catch these, report them as info-level logs in the `rails_events category` through the newly added `Appsignal::Integrations::ActiveSupportEventReporter`.

### DIFF
--- a/spec/lib/appsignal/cli/diagnose_spec.rb
+++ b/spec/lib/appsignal/cli/diagnose_spec.rb
@@ -490,18 +490,18 @@ describe Appsignal::CLI::Diagnose, :api_stub => true, :send_report => :yes_cli_i
             "Agent diagnostics",
             "  Error while parsing agent diagnostics report:",
             "    Output: invalid agent\njson"
-          # Ruby 3.x: "unexpected token at 'invalid'"
+          # Ruby 3.x / JRuby: "unexpected token at 'invalid...'"
           # Ruby 4.x: "unexpected character: 'invalid'" or "source is not valid JSON!"
           error_pattern =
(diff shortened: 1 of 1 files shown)


spec/lib/appsignal/hooks/active_support_event_reporter_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+          def initialize
+            @subscribers = []
+          end
+
+          def subscribe(subscriber)
+            @subscribers << subscriber
+          end
+        end)
+
+        mock_event_reporter = MockEventReporter.new
```
tombruijn: I recommend using an instance double for this, so that if the Rails API for this ever changes, our tests fail too, and it doesn't silently keep passing the test. Keeping us unaware anything is broken.
https://rspec.info/features/3-12/rspec-mocks/verifying-doubles/instance-doubles/

lib/appsignal/config.rb
```diff
@@ -102,6 +102,7 @@ def dsl_config_file?
       :enable_gvl_global_timer => true,
       :enable_gvl_waiting_threads => true,
       :enable_rails_error_reporter => true,
+      :enable_active_support_event_reporter => true,
```
tombruijn: Nit: maybe we should make the option specific for logging in case we want to add another reporter that does add events to the transaction events later?

```suggestion
      :enable_active_support_event_log_reporter => true,
```

PR #1499, opened 2026-03-04
Respect ActiveJob log_arguments configuration

### LINKED ISSUE #1487: Respect log_arguments = false in ActiveJob
AppSignal currently collects `ActiveJob` arguments for error and performance events even when a job is configured with `log_arguments = false`. This can cause sensitive data to appear in AppSignal despite Rails being explicitly configured not to log them.

This enhancement proposes that AppSignal respects `log_arguments = false` when instrumenting `ActiveJob`, both when it is configured globally (e.g. on `ApplicationJob`) and per job class.

This is problematic for jobs like `ActionMailer::MailDeliveryJob` and DeviseMailers as they could use plain string arguments, which cannot be filtered using `filter_parameters`.

[Intercom](https://app.intercom.com/a/inbox/yzor8gyw/inbox/conversation/215472579651940?view=List)

### Proposed behavior
When instrumenting ActiveJob, we should check the job’s effective `log_arguments` setting (including inherited values). If `job.class.log_arguments == false`, job arguments should not be collected or reported (or should be replaced with `[FILTERED]`)

This behavior should apply:
- when `log_arguments` is set globally on `ApplicationJob`
- when it is set on individual `ActiveJob` classes

### Acceptance criteria:

- [ ] ActiveJob arguments are not collected or displayed when log_arguments = false
- [ ] Both global (ApplicationJob) and per-job configurations are respected

jvanbaarsen:
When an ActiveJob class has log_arguments set to false, AppSignal now skips collecting job arguments for that transaction. This prevents sensitive data from appearing in AppSignal when Rails is explicitly configured not to log them.

This applies when log_arguments is set:
- Directly on a job class (e.g. self.log_arguments = false)
- On a parent class (e.g. ApplicationJob), inherited by subclasses

The implementation adds a log_arguments? helper that resolves the job class from the serialized job data and checks the class-level log_arguments? setting. It defaults to true (collect arguments) when the class cannot be resolved or does not support the method, ensuring backward compatibility.

Closes #1487

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -180,6 +180,27 @@ rake build_matrix:github:generate
 
 The script to generate the `.github/workflows/ci.yml` file is part of the `Rakefile`.
 
+## Commit Message Convention
+
+This project uses [Lintje](https://lintje.dev/) to lint commit messages in CI. All commits must follow these rules:
+
+- **Subject line**: Maximum 50 characters. Use imperative mood (e.g. "Fix bug" not "Fixed bug").
+- **Message body**: Required. Add a blank line after the subject, then describe what changed and why. Wrap lines at 72 characters.
+- **Ticket/issue reference**: Include a reference to a GitHub issue in the message body (e.g. `Fixes #1487` or `Part of #1487`).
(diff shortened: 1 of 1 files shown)


tombruijn:
Closing in favor of #1507

PR #1515, opened 2026-05-22
Support OpenTelemetry via collector mode

unflxw:
**This is a work in progress. Nothing about this approach is set in stone.**

This is a starting point for testing out the different potential approaches to OpenTelemetry support via the (hosted) collector on the Ruby integration.

It is in part a re-implementation of the work done in the Python integration, particularly of [Python's experimental collector mode](https://github.com/appsignal/appsignal-python/pull/247) and [Python's metric helpers](https://github.com/appsignal/appsignal-python/pull/127).

## What this does

- Defines a `collector_endpoint` configuration option.
- When the config option is present and the OpenTelemetry dependencies are installed, initialises an OpenTelemetry SDK with traces, metrics and logs pointing at that endpoint.
- Defines other configuration options that are used to tell the collector how to process and filter data, and sends those configuration options via the OpenTelemetry resource.
- Routes metrics emitted via AppSignal's metric helpers, and logs emitted via `Appsignal::Logger`, through OpenTelemetry.
- Transforms transactions and events emitted via `Appsignal::Transaction` into OpenTelemetry span trees and emits them through OpenTelemetry, testing each library and framework integration thoroughly.

## What this does not do (yet)

- Change how the agent runs in any way. It still runs, even if the collector endpoint is set. It does nothing other than sending host metrics.
- Configure any of OpenTelemetry's own instrumentations.

## Non-decisions

The following are things that the current implementation does, but that are explicitly not set in stone:

- **The OpenTelemetry SDK for metrics and logs in Ruby is experimental** -- logs more so than metrics. We may decide this is fine _(we use the experimental Python one for logs in the Python integration, and we've not ran into any issues so far)_ or we may decide it's not. For logs, we get trace IDs and span IDs for correlation, so there's value in using OpenTelemetry to route them. For metrics, not so much, as they get AppSignal-ified on the OpenTelemetry pipeline too.
- **The agent still runs.** We could leave it running, which would mean we don't lose host metrics or drop support for NGINX or StatsD. This might actually be a good approach _(we could even backport it into the Python integration, if we like it!)_ but it should be an explicit decision. We could potentially put it behind a config flag, or derive whether it's initialised at all from whether any config flags require it.
- We should take a double look at the config options. The warnings that exist for the Python integration about config options that do nothing in collector mode have been replicated, but for example, we should have `enable_opentelemetry_http` emit a warning as well when used alongside collector mode, as that's likely not intended behaviour either.


lib/appsignal/opentelemetry.rb
```diff
@@ -0,0 +1,160 @@
...
+
+        attrs = {
+          "appsignal.config.name" => config[:name],
+          "appsignal.config.environment" => config.env,
+          "appsignal.config.push_api_key" => config[:push_api_key],
+          "appsignal.config.revision" => revision,
+          "appsignal.config.language_integration" => "ruby",
+          "service.name" => service_name,
+          "host.name" => host_name,
+          "appsignal.service.process_id" => Process.pid,
```
tombruijn: We haven't added/documented a config option like this yet. What would it do?

tombruijn (commented):
I skimmed it. So far so good :)

lib/appsignal/opentelemetry.rb
unflxw: Neat little rabbit hole. Long story short, as far as I can tell, nothing.

(Note that this is not a config option, it's a resource attribute we're sending to the agent/collector)

This was lifted from the Python integration, [which sends it as well](https://github.com/appsignal/appsignal-python/blob/0650df9ba096221d40b38f566f87f32772032e2e/src/appsignal/opentelemetry.py#L299), but I cannot for the life of me recall why we put it there. [Git history says you might know](https://github.com/appsignal/appsignal-python/commit/1a5466a7a86fa67a76b7d64e6dfa08fafebb86b3), though I can't make sense of it from the commit or the associated PR. I can't find references to it in either the agent or the collector.

Happy to remove it.

tombruijn: I think it's jostling my memory a little bit. It's probably one of those things that we added early so that when we would implement a process monitoring for the Python integration in the agent, we would already have the config option. We never got to implementing that in the Python package though.

We can remove it for now. It's not like the Ruby gem with this setup (hosted collector) would talk to something that could locally instrument the process.
But maybe in the future... 😅

unflxw: Will remove then! That future version of us that needs it for something can add it back.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,206 @@
...
+      end
+
+      def set_namespace(namespace) # rubocop:disable Naming/AccessorMethodName
+        # Only the attribute can change here: SpanKind is fixed at span
+        # creation (immutable in OTel) from the initial namespace. A later
+        # namespace override updates `appsignal.namespace` but not the kind --
+        # the collector uses the attribute for the namespace and the kind only
+        # to pick the subtrace root, so this is fine for the rare late change.
+        @span.set_attribute("appsignal.namespace", namespace)
+      end
```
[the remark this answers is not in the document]
unflxw: (should be fine in practice because transactions will not duplicate, but let's fix it)

unflxw:
Closing in favour of #1535.

PR #1516, opened 2026-05-27
Skip set_error when Active Job transaction is nil

jbockler:
We hit this in production when stopping Resque workers: AppSignal reported `NoMethodError: undefined method 'set_error' for nil` from `Appsignal::Hooks::ActiveJobHook::ActiveJobClassInstrumentation#execute`, masking the underlying `Resque::TermException: SIGTERM` that triggered worker shutdown.

The cause: when an async signal unwinds through `execute` before its `transaction = …` assignment completes, the local `transaction` stays at its parser-default `nil`. The `rescue Exception` clause then calls `nil.set_error(exception)`, raising the `NoMethodError` on top of the original `SignalException`.

### DIFF
--- a/.changesets/fix-active-job-nil-transaction-set-error.md
+++ b/.changesets/fix-active-job-nil-transaction-set-error.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+Prevent a `NoMethodError` in the Active Job hook when a job is interrupted by `SIGTERM`.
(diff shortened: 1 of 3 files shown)


tombruijn:
Hi @jbockler, thank you for the PR! I can see how this can happen in the rare circumstance where the job is exited before the transaction variable is assigned.

While this would fix it, I have chosen to implement the fix in #1517 by wrapping the code in a more explicit begin-rescue-ensure block rather than anticipate that the transaction variable can be `nil`. We have a different strategy for that, using the `NilTransaction` class, that avoids us from having to nil checks all the time.

I will close this PR in favor of #1517.

jbockler:
Alright, thanks for looking into this @tombruijn !

tombruijn:
I published the fix in Ruby gem 4.8.5.

PR #1517, opened 2026-06-01
Avoid transactions being nil in instrumentations

tombruijn:
Fix the instrumentations that use a rescue and/or ensure block for the entire method when we instrument other libraries by ensuring that the transaction variable is always assigned before those blocks can be called. Which means defining the variable outside of the begin-block.

This avoids the situation where the Ruby process is told to shut down before the transaction is being created in the instrumentation method. Which is the only point when the transaction variable can be `nil` and then raise a `NoMethodError`.

We can guarantee the transaction is then never `nil` because even if the extension is not installed/active, and the transaction can't be created, we return a `NilTransaction` class which doesn't cause the `NoMethodError` to be raised.

Closes #1516

### DIFF
--- a/lib/appsignal/hooks/active_job.rb
+++ b/lib/appsignal/hooks/active_job.rb
@@ -66,49 +66,51 @@ module Appsignal
               Appsignal::Transaction.create(Appsignal::Transaction::BACKGROUND_JOB)
             end
 
-          if transaction
-            transaction.add_params_if_nil(job["arguments"])
+          begin
+            if transaction
+              transaction.add_params_if_nil(job["arguments"])
 
-            transaction_tags = ActiveJobHelpers.transaction_tags_for(job)
-            transaction.add_tags(transaction_tags)
+              transaction_tags = ActiveJobHelpers.transaction_tags_for(job)
+              transaction.add_tags(transaction_tags)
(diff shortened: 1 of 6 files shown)


lipskis (approved):

lib/appsignal/hooks/active_job.rb
```diff
@@ -66,49 +66,51 @@ def execute(job) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/Perceiv
               Appsignal::Transaction.create(Appsignal::Transaction::BACKGROUND_JOB)
             end
 
-          if transaction
-            transaction.add_params_if_nil(job["arguments"])
+          begin
+            if transaction
```
unflxw: I think this should be the other way around -- `if transaction` before `begin`, not after:

```ruby
if transaction
  begin
    # ...
  rescue
    # do something with transaction
  end
end
```

unflxw (approved):

lib/appsignal/hooks/active_job.rb
tombruijn: Actually, I think the if-statement can be removed. It's always Transaction or NilTransaction, not nil.

PR #1522, opened 2026-06-26
Instrument Que job enqueues

unflxw:

This PR does the preliminary changes to the Que integration needed in order to be able to support this behaviour. 

---

Record an `enqueue.que` event when a Que job is enqueued, and a `bulk_enqueue.que` event for the whole batch on a bulk enqueue, so enqueues show up in the transaction timeline. The bulk wrapper records one event and its inner enqueues pass through.

The bulk module is only prepended on Que 2+, which has `bulk_enqueue`, so no phantom method is defined on Que 1.

### DIFF
--- a/lib/appsignal/integrations/que.rb
+++ b/lib/appsignal/integrations/que.rb
@@ -43,7 +43,7 @@ module Appsignal
     # from within a web request or another job); otherwise it's a transparent
     # pass-through.
     module QueClientPlugin
-      def enqueue(*_args, **_rest)
+      def enqueue(*_args, job_options: {}, **_rest)
         # Inside a Que 2 `bulk_enqueue` block the batch is recorded once by the
         # `bulk_enqueue` wrapper, so each inner enqueue is a pass-through to
         # avoid recording an event per job.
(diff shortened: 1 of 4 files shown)


lib/appsignal/integrations/que.rb
```diff
@@ -34,5 +34,41 @@ def _run(*args)
...
+        # avoid recording an event per job.
+        return super if bulk_insert_in_progress?
+
+        Appsignal.instrument("enqueue.que") { super }
+      end
+
+      private
+
+      def bulk_insert_in_progress?
+        !Thread.current[:que_jobs_to_bulk_insert].nil?
```
tombruijn:
```suggestion
        # Set by the Que gem
        # https://github.com/que-rb/que/blob/caa8b25cdd9da7fa74d27532ad6acf0a8c23f115/lib/que/job.rb#L116
        !Thread.current[:que_jobs_to_bulk_insert].nil?
```

tombruijn (approved):

lib/appsignal/integrations/que.rb
unflxw: Good suggestion, but rather than document Que's thread-local I've stopped reading it altogether: the bulk_enqueue wrapper now sets our own flag (which also addresses Copilot's comments), so there's no Que internal left to link to.

PR #1523, opened 2026-06-26
Improve Faraday support

unflxw:
Part of https://github.com/appsignal/appsignal-ruby/pull/1515, spun as a separate PR to make reviewing easier.

This work is part of ensuring that distributed tracing has a dedicated span to do context propagation from, following the OpenTelemetry semantic conventions for HTTP clients: a `CLIENT` span is created when the HTTP request is performed, the trace context is propagated through the request's headers, and the server receiving that request continues the trace with a `SERVER` span, which is a child of the `CLIENT` span.

This PR does the preliminary improvements to the Faraday integration needed in order to be able to support this behaviour.

In this case, those changes are quite substantial: our pre-existing "Faraday integration" was an undocumented `ActiveSupport::Notifications` event formatter that relied on the customer manually configuring some Faraday instrumentation middleware. This PR turns this pre-existing pseudo-support into a more proper automated integration, matching those for our other HTTP libraries. It still relies in the same underlying notifications event formatter, so it has ActiveSupport as a dependency as a result. This could be changed -- the intent was to limit the change surface.

It also changes our Faraday event formatter to follow the format of our Net::HTTP event formatter. As Faraday is a wrapper for other HTTP libraries, it also ensures the underlying HTTP library's own instrumentation (Net::HTTP or Excon) is disabled when used.

---

### [Add the instrument_faraday config option](https://github.com/appsignal/appsignal-ruby/commit/6d82ea695b2880335742dbf7409cab9bd6e21ecd)

Add the option that toggles Faraday instrumentation. It defaults to on
and can be disabled with the APPSIGNAL_INSTRUMENT_FARADAY environment
variable.

### [Add a Faraday integration](https://github.com/appsignal/appsignal-ruby/commit/b712aaf0d3a489012496bc69dab79b4179b7d4a7)

Instrument Faraday requests automatically. The integration installs
Faraday's own instrumentation middleware onto every connection, so a
request.faraday event is recorded without the user adding the
middleware themselves. Tested against Faraday 1 and 2.

The event title leaves out the request path, to match the Net::HTTP
integration.

### [Suppress downstream HTTP under Faraday](https://github.com/appsignal/appsignal-ruby/commit/702f1e7a2db405637e22f95f7172102457b358aa)

Faraday runs on a downstream HTTP client that AppSignal may also
instrument -- Net::HTTP (the default adapter) or Excon. Flag the
transaction while Faraday handles a request so that client skips its
own event, and the request is recorded once, as the Faraday event,
instead of as nested client events.

### [Update diagnose tests for instrument_faraday](https://github.com/appsignal/appsignal-ruby/commit/b2c8684ceb85b1404d8ae2c47570a92059e780f4)

The diagnose output lists every config option, so add the new
instrument_faraday option to the diagnose test expectations.


lib/appsignal/transaction.rb
```diff
@@ -290,6 +290,24 @@ def store(key)
...
+    # Run a block during which downstream HTTP client integrations (Net::HTTP,
+    # ...) skip recording their own event. Used when an outer HTTP client
+    # integration (Faraday) already records the request, so the same request is
+    # not instrumented twice as nested client events.
+    def suppress_http_client_events
+      store("http_client")[:suppressed] = true
+      yield
+    ensure
+      store("http_client")[:suppressed] = false
+    end
```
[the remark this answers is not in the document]
unflxw: Not a concern at the moment since only Faraday does this, but worth adding this now and not leaving the footgun around.

lib/appsignal/event_formatter/faraday/request_formatter.rb
tombruijn, line 1: You mentioned that this keeps a dependency on active support. Could we remove that dependency? What if we update the Faraday instrumentation to emit its own event by wrapping the app.call with an Appsignal.instrument? 

We don't have to do that in this PR but we could make another one to make that change.

tombruijn (approved):

lib/appsignal/event_formatter/faraday/request_formatter.rb
unflxw: Yeah, the framing for this work was initially to "grandfather" the existing Faraday "instrumentation" (the `ActiveSupport::Notifications` event formatter) into a real Faraday instrumentation, hence why it keeps its dependency in the event formatter, and therefore in the underlying `ActiveSupport::Notifications` event.

That dependency is entirely vestigial now, since we do already wrap `@app.call` for other purposes, we might as well call `Appsignal.instrument` there and drop the formatter path entirely.

PR #1525, opened 2026-06-26
Instrument Resque job enqueues

unflxw:

This PR does the preliminary changes to the Resque integration needed in order to be able to support this behaviour.

---

Prepend onto the Resque singleton so each enqueue through `Resque.push` records an `enqueue.resque` event on the active transaction, so enqueues made from within a web request or another job show up in the event timeline. Mirrors the Sidekiq, Que and Shoryuken enqueue instrumentation.

Like any AppSignal event, it only records when a transaction is active; an enqueue with no transaction is a transparent pass-through.

Backport of the enqueue instrumentation from the OpenTelemetry branch, without the trace-context propagation, which is collector-mode only.

### DIFF
--- a/lib/appsignal/integrations/resque.rb
+++ b/lib/appsignal/integrations/resque.rb
@@ -34,13 +34,13 @@ module Appsignal
     #
     # @!visibility private
     module ResquePushIntegration
-      def push(_queue, _item)
+      def push(_queue, item)
         # Under Active Job the enqueue is already recorded as an
         # `enqueue.active_job` event, so skip recording it again here.
         return super if Appsignal::Transaction.current? &&
           Appsignal::Transaction.current.job_enqueue_events_suppressed?
 
-        Appsignal.instrument("enqueue.resque") { super }
+        Appsignal.instrument("enqueue.resque", "enqueue #{item["class"]} job") { super }
       end
     end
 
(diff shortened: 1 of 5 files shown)


lib/appsignal/integrations/resque.rb
```diff
@@ -25,6 +25,20 @@ def perform
...
+    # shows up under the active transaction.
+    #
+    # Like all AppSignal events, this only records when there's an active
+    # transaction (e.g. enqueuing from within a web request or another job).
+    # An enqueue with no transaction is a transparent pass-through.
+    #
+    # @!visibility private
+    module ResquePushIntegration
+      def push(_queue, _item)
+        Appsignal.instrument("enqueue.resque") { super }
```
tombruijn: For the Ruby PR could we also include the worker or job name in the title or body of the event here?

tombruijn (approved):

lib/appsignal/integrations/resque.rb
unflxw: Implemented across all background job enqueue PRs (https://github.com/appsignal/appsignal-ruby/pull/1527, https://github.com/appsignal/appsignal-ruby/pull/1524, https://github.com/appsignal/appsignal-ruby/pull/1525, https://github.com/appsignal/appsignal-ruby/pull/1522, https://github.com/appsignal/appsignal-ruby/pull/1521) along with functionality to prevent duplicate nested events from showing up on ActiveJob-over-SomeOtherBackgroundJobLibrary scenarios.

PR #1527, opened 2026-06-26
Refactor Active Job enqueues

unflxw:

This PR does the preliminary change to the Active Job integration needed in order to support this behaviour, without the trace-context propagation, which is collector-mode only.

---

Unlike the other job integrations, Active Job already records an `enqueue.active_job` event in agent mode: it rides on Rails' native `enqueue.active_job` notification, which the generic `ActiveSupport::Notifications` path picks up. This PR moves that recording into AppSignal's own instrumentation. A prepend on `ActiveJob::Base#enqueue` records the event, and the notifications path now suppresses the native one so the enqueue is recorded exactly once.

There is no user-facing change: the same single `enqueue.active_job` event still shows up on the active transaction when enqueuing from within a web request or another job. The point is to give AppSignal a single enqueue event it owns -- the seam the collector-mode work later hangs trace-context propagation off of -- and to land that instrumentation refactor on its own, so it can be reviewed and shipped independently of the OpenTelemetry changes.

### DIFF
--- a/lib/appsignal/hooks/active_job.rb
+++ b/lib/appsignal/hooks/active_job.rb
@@ -56,7 +56,7 @@ module Appsignal
       # @!visibility private
       module ActiveJobEnqueueInstrumentation
         def enqueue(*, **)
-          Appsignal.instrument("enqueue.active_job") do
+          Appsignal.instrument("enqueue.active_job", "enqueue #{self.class.name} job") do
             # Active Job enqueues through an adapter (Sidekiq, Resque, ...) that
             # has its own enqueue instrumentation. Suppress it so the enqueue is
             # recorded once, as this event, rather than as nested Active Job +
--- a/lib/appsignal/hooks/active_job.rb
+++ b/lib/appsignal/hooks/active_job.rb
@@ -29,6 +29,8 @@ module Appsignal
         ActiveSupport.on_load(:active_job) do
           ::ActiveJob::Base
             .extend ::Appsignal::Hooks::ActiveJobHook::ActiveJobClassInstrumentation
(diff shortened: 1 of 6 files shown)


lib/appsignal/hooks/active_job.rb
```diff
@@ -41,6 +43,23 @@ def install
...
+      # Wrapping `enqueue` ourselves -- rather than relying on Rails' native
+      # `enqueue.active_job` notification, which the AppSignal notifications
+      # path now suppresses -- gives us a single event we own. Like all
+      # AppSignal events, this only records when there's an active transaction;
+      # an enqueue with no transaction is a transparent pass-through.
+      #
+      # @!visibility private
+      module ActiveJobEnqueueInstrumentation
+        def enqueue(*, **)
+          Appsignal.instrument("enqueue.active_job") { super }
```
tombruijn: Could we include the worker name here so that people can see what kind of job is being enqueued?

tombruijn (approved):

lib/appsignal/hooks/active_job.rb
unflxw: Implemented across all background job enqueue PRs (https://github.com/appsignal/appsignal-ruby/pull/1527, https://github.com/appsignal/appsignal-ruby/pull/1524, https://github.com/appsignal/appsignal-ruby/pull/1525, https://github.com/appsignal/appsignal-ruby/pull/1522, https://github.com/appsignal/appsignal-ruby/pull/1521) along with functionality to prevent duplicate nested events from showing up on ActiveJob-over-SomeOtherBackgroundJobLibrary scenarios.

PR #1535, opened 2026-07-08
OpenTelemetry collector mode

unflxw:
Somewhat reviewable re-do of #1515.

While it's still a massive amount of changes, it separates the signal (implementing collector mode for metrics, traces and logs, as well as adding trace context propagation to our existing library and framework integrations) from the noise (making each and every integration's existing tests run both in agent mode and collector mode)

The sections below follow the branch in commit order. Where a run of commits was already reviewed in its own pull request, the section links to that pull request and lists the commits it covers, rather than repeating messages that have already been read. The remaining sections are commits that have had no pull request of their own.

### [Add OpenTelemetry collector mode](https://github.com/appsignal/appsignal-ruby/commit/84d83b657d65a93859ba73f295216a3b0ee9cab1)

Setting `collector_endpoint`, or `APPSIGNAL_COLLECTOR_ENDPOINT`, puts
the integration in collector mode. `Appsignal.start` then boots an
OpenTelemetry SDK that exports OTLP over HTTP to that endpoint,
alongside the usual C extension agent.

The OpenTelemetry gems are not gemspec dependencies, because collector
mode is optional. The boot checks for them and falls back to the agent
when they are absent or too old.

### [Emit OpenTelemetry logs in collector mode](https://github.com/appsignal/appsignal-ruby/commit/db42338b93d34075038a4c0f4c3cc71c1f446cc8)

`Appsignal::Logger` goes through a backend. The extension backend
keeps the existing C extension logging, and an OpenTelemetry backend
emits log records through the SDK's logger provider in collector mode.

Ruby severities and formats are mapped to their OTLP equivalents.
`appsignal.group` and `appsignal.format` are hard overrides, so a user
attribute cannot spoof them.

### [Emit OpenTelemetry metrics in collector mode](https://github.com/appsignal/appsignal-ruby/commit/bc536ae2c2c1de6744fec54eff0cd04fd7a2c730)

The custom metric helpers go through a backend. The extension backend
keeps the C extension path, including its out-of-range warning, and an
OpenTelemetry backend records gauges, up/down counters and histograms
through the SDK's meter provider in collector mode.

With the metrics and logger backends both in place, `Appsignal.start`
can fall back to the agent when the OpenTelemetry SDK fails to boot, so
telemetry is never silently dropped.

### [Test the metric probes in collector mode](https://github.com/appsignal/appsignal-ruby/commit/cb1d004bbcff1f53881c82476ef58d0bb32b1d4b)

The probes emit through the metric helpers, so they need no changes of
their own. Running each metric example in both modes proves their
output routes to OpenTelemetry when collector mode is active.

### [Emit OpenTelemetry traces in collector mode](https://github.com/appsignal/appsignal-ruby/commit/0b714837e058e3b4c6c456a1287c266c5cae6b0d)

`Appsignal::Transaction` goes through a backend. The extension backend
keeps the existing C extension behaviour, and now owns the sample data
serialization, the breadcrumb buffering and the agent's error cause
shape. An OpenTelemetry backend maps a transaction to a root span and
each event to a child span.

One `Transaction` drives two error models, so the backend answers
`records_errors_eagerly?`. The extension collects errors and then
duplicates the transaction once per error at completion. The
OpenTelemetry backend records each error as an `exception` span event
straight away.

Three mappings decide the shape of the exported trace:

- A span's kind comes from the namespace. A web or Action Cable
  transaction is SERVER, and a background job is CONSUMER. A datastore
  client event passes `opentelemetry_kind: :client`. The kind is set at
  creation, because it cannot change afterwards.
- An event span is named `title || name`, and the event name is kept as
  the `appsignal.category` attribute. A SQL body maps to
  `db.query.text`, and every other body to `appsignal.body`.
- An error attaches to the currently open span, with its cause chain in
  one `appsignal.error_causes` JSON attribute. A discarded transaction
  sets `appsignal.ignore_subtrace`, which tells the collector to drop
  it.

### [Test integrations in collector mode](https://github.com/appsignal/appsignal-ruby/commit/9a611e9be9b19aaab3bb696df72212cf8a642244)

Every existing event-emission example in the hook, integration, Rack
and loader specs now runs in both agent and collector mode. The
collector-mode half asserts the span shape: its name, category, body,
kind and parent. No library code changes, so this exercises the trace
backend through every integration.

### [Inject trace context into outgoing HTTP requests](https://github.com/appsignal/appsignal-ruby/commit/6ce933f7ac8741423eba5b816291f6cfb4a21bc7)

In collector mode the HTTP client integrations write the current W3C
trace context onto their outgoing requests, so the service being called
continues the same trace. Net::HTTP, HTTP.rb, Excon and Faraday each
inject a `traceparent` header and mark the request span with CLIENT
kind. HTTP.rb injects on every hop, and Excon injects through a
middleware.

Injection is a no-op outside collector mode.

### [Extract trace context from incoming web requests](https://github.com/appsignal/appsignal-ruby/commit/37f577c264adf332e8215b8d50b17866034add77)

In collector mode the Rack middleware, the Rack event handler and the
Webmachine integration read an incoming W3C `traceparent` header and
pass it to the transaction. A web transaction continues under the
remote span. A job transaction starts its own trace and links back to
it. With no context, or outside collector mode, the root span stands
alone.

### [Propagate trace context across background jobs](https://github.com/appsignal/appsignal-ruby/commit/dc081b6e3da6fef4aa5819de059cb16137cef290)

In collector mode each job library carries the W3C trace context across
the enqueue and perform boundary. On enqueue the client writes the
current context onto the job payload and marks the enqueue span with
PRODUCER kind. On perform the server side reads it back and links the
job's trace to the enqueuing span.

Each library carries the context on its own channel, such as a job hash
key, an SQS message attribute, a Que tag, or Active Job's serialized
arguments. When a channel is full the propagation is skipped rather
than the enqueue broken.

### [Run error blocks eagerly in collector mode](https://github.com/appsignal/appsignal-ruby/commit/4907792e89e70e0c8d141930371c5b5240ca671d)

In collector mode an error is recorded when it is added, but its block
was held until completion. Deferral only makes sense in agent mode,
where blocks run against the duplicate transactions fabricated at
completion. The block now runs when the error is added, so side effects
that target the current span, such as breadcrumbs or nested errors,
land where the error was reported rather than on the root span.

`@error_blocks` was carrying two jobs: the set of distinct errors, which
drives `last_errors`, the dedup check and the error limit, and the
per-error blocks that only agent mode runs. `@errors` is now the
distinct-error set used in both modes, and `@error_blocks` holds only
blocks and is populated only in agent mode.

`records_errors_eagerly?` is renamed to `supports_multiple_errors?`.
The question at the call site is whether the backend can hold more than
one error on a transaction, not how it records them.

### [Mark Sequel query spans as CLIENT kind in collector mode (#1546)](https://github.com/appsignal/appsignal-ruby/pull/1546)

- [Tag Sequel query spans CLIENT in collector mode](https://github.com/appsignal/appsignal-ruby/commit/5e1d1cbdf7c34595393bda3cd98268300bb6e49d)

### [Extract trace context onto an empty context (#1542)](https://github.com/appsignal/appsignal-ruby/pull/1542)

- [Extract trace context onto an empty context](https://github.com/appsignal/appsignal-ruby/commit/ecfb15fe9f6aa55e74c695c13e26c38462d0b864)

### [Guard blocks handed to the transaction (#1544)](https://github.com/appsignal/appsignal-ruby/pull/1544)

- [Guard blocks handed to the transaction](https://github.com/appsignal/appsignal-ruby/commit/10e49ff19aec3145b0828494e9a4173cc4c5e3ea)

### [Drop actionless transactions in collector mode (#1541)](https://github.com/appsignal/appsignal-ruby/pull/1541)

- [Drop actionless transactions in collector mode](https://github.com/appsignal/appsignal-ruby/commit/42e6b67a4049369e93d73a2b63ceb5578f50a561)

### [Record the Delayed Job enqueue as a producer span (#1538)](https://github.com/appsignal/appsignal-ruby/pull/1538)

- [Record the Delayed Job enqueue as a producer span](https://github.com/appsignal/appsignal-ruby/commit/f1ad6b4af8ceb0de7efb62b0718966e65dff1215)

### [Dual-mode transaction and helper spec coverage (#1547)](https://github.com/appsignal/appsignal-ruby/pull/1547)

- [Dual-mode transaction and helper spec coverage](https://github.com/appsignal/appsignal-ruby/commit/d5ac552ad150d4fd6b4c4ec4c6d810af06e9b7ba)

### [Tag integration spans with instrumentation scope (#1551)](https://github.com/appsignal/appsignal-ruby/pull/1551)

- [Resolve OTel logger per emit, drop the cache](https://github.com/appsignal/appsignal-ruby/commit/383b34c93898a98eb9faab3ff3c971566f05c76c)
- [Add opentelemetry_scope to instrumentation](https://github.com/appsignal/appsignal-ruby/commit/467ec4b122473180f3eb23bedb15066704470c1d)
- [Derive AS::N span scope from the event group](https://github.com/appsignal/appsignal-ruby/commit/caced5c347a3b17314fcdfc1747e560628bca83c)
- [Test AS::N derived instrumentation scope](https://github.com/appsignal/appsignal-ruby/commit/ad024aaf18c6315d852d2aad9491c3670735e50b)
- [Tag Rack and web framework spans with scope](https://github.com/appsignal/appsignal-ruby/commit/9fad25b5bbe55b9aadee8b77e7bdfdac44ad9e79)
- [Test Rack and web framework span scope](https://github.com/appsignal/appsignal-ruby/commit/b28e289823752706b79e22ba1d546ae88070f32f)
- [Tag remaining integration spans with scope](https://github.com/appsignal/appsignal-ruby/commit/d9c57a1e2c39ec123308c3157dab25f9d25aa6bd)
- [Test remaining integration scopes](https://github.com/appsignal/appsignal-ruby/commit/07a56bd48a2a29a435692c416b1078e8abe7837c)

### [Add the appsignal-opentelemetry companion gem (#1554)](https://github.com/appsignal/appsignal-ruby/pull/1554)

- [Add the appsignal-opentelemetry companion gem](https://github.com/appsignal/appsignal-ruby/commit/9b122f6fcd04dba1e6d1881cac50f29325a90fe4)
- [Bound collector OpenTelemetry gem versions](https://github.com/appsignal/appsignal-ruby/commit/e5a91ac7b66dc7efbb1685d7f5f0dcd609acad83)
- [Install package deps in build:packages task](https://github.com/appsignal/appsignal-ruby/commit/2dfc1eb8ebdbde9f7f01ae7d4286bb142daf4162)

### [Report allocation counts in collector mode (#1550)](https://github.com/appsignal/appsignal-ruby/pull/1550)

- [Don't emit metrics for actionless transactions](https://github.com/appsignal/appsignal-ruby/commit/082782da912997b3e0dc243ac4850fdfbf3f373f)
- [Report allocation counts in collector mode](https://github.com/appsignal/appsignal-ruby/commit/1d522b97727205f955f1cc07e777ee1a7b5193b4)
- [Use a clearer placeholder for unfinished events](https://github.com/appsignal/appsignal-ruby/commit/2cabd150d8c88b28a213f153115deacab27d58dd)

### [Set transaction OpenTelemetry kind and relationship explicitly (#1552)](https://github.com/appsignal/appsignal-ruby/pull/1552)

- [Set transaction span kind and relationship](https://github.com/appsignal/appsignal-ruby/commit/7e227c8f9b06a3fcd54550989dd08959fd5f350d)
- [Split params into request and function channels](https://github.com/appsignal/appsignal-ruby/commit/c9898e2470399ee324b11144310b118b70079089)
- [Parent background job spans to the enqueuer](https://github.com/appsignal/appsignal-ruby/commit/0ef05280be2adb86583b6a9b48d2c2c2e1552345)

### [Document the OpenTelemetry keyword arguments](https://github.com/appsignal/appsignal-ruby/commit/4d693e287354a900933fdf050c197327e465faf0)

`opentelemetry_kind`, `opentelemetry_relationship` and
`opentelemetry_context` carried documentation, but `opentelemetry_scope`
did not, and neither did `opentelemetry_kind` on `Appsignal.instrument`.
The type signatures are generated from these comments, so the
undocumented arguments were generated as `untyped`. They are documented
everywhere they are accepted now, so `opentelemetry_scope` generates as
a `[String, String]` pair. The signatures are regenerated to match.

### [Put event category in collector span name (#1553)](https://github.com/appsignal/appsignal-ruby/pull/1553)

- [Put event category in collector span name](https://github.com/appsignal/appsignal-ruby/commit/4e9d5dc9e085d724b102348c353340f911d05f2b)
- [Rename the event span name writer](https://github.com/appsignal/appsignal-ruby/commit/5c7d3245e6cc03b675fb87546908983d1526947d)

### [Link bulk enqueued Que jobs instead of parenting](https://github.com/appsignal/appsignal-ruby/commit/10a7f307fce63ad709a0ca91fe5328f9256833a9)

Every job enqueued by `Que.bulk_enqueue` shares one producer span,
because the whole batch is recorded as a single `bulk_enqueue.que`
event. A span can have only one parent, so parenting hangs the entire
batch off that one span. The OpenTelemetry messaging conventions ask
for links by default, and allow the producer to be the parent only when
it produced a single message.

Que records nothing that tells its two enqueue paths apart. Both insert
paths write the same columns, and a job's `data` column holds only its
tags. The enqueue side marks a batch with a tag, and the worker links
instead of parenting when it finds that tag. The tag contains no colon,
so a reader that splits tags on the colon to rebuild the trace context
carrier ignores it.

Que allows five tags per job and the trace context already takes one.
On the batch path the marker and the trace context are kept or dropped
together, because propagating without the marker would make the worker
parent the whole batch. When there is no room for both, the batch is
enqueued without trace context.

### [Trim shared frames from error cause backtraces (#1559)](https://github.com/appsignal/appsignal-ruby/pull/1559)

- [Trim shared frames from error cause backtraces](https://github.com/appsignal/appsignal-ruby/commit/1872a71e3bbcb1d249acfbdaca4f4f6ae753fb26)
- [Report how many cause lines were omitted](https://github.com/appsignal/appsignal-ruby/commit/95a237b63e2ac45c0bbd2e60dba287f8db3dc7ad)

### [Skip trace context when enqueue events are off (#1561)](https://github.com/appsignal/appsignal-ruby/pull/1561)

- [Skip trace context when enqueue events are off](https://github.com/appsignal/appsignal-ruby/commit/6faf76f28971ad6d2ed9a823bf89e34fa0fe02b1)

### [Emit OpenTelemetry attributes for integrations (#1562)](https://github.com/appsignal/appsignal-ruby/pull/1562)

- [Add an OpenTelemetry span attributes method](https://github.com/appsignal/appsignal-ruby/commit/d2c3ccca594cd4ec0ac9bcf95cfec2252a1b9d88)
- [Describe outgoing HTTP requests](https://github.com/appsignal/appsignal-ruby/commit/5c006c876c36c0ae1d5ebc671a9ebb92e7751297)
- [Describe incoming HTTP requests](https://github.com/appsignal/appsignal-ruby/commit/39bc0eb1345a64345143cbe2123da59839762e8b)
- [Describe database queries](https://github.com/appsignal/appsignal-ruby/commit/6c4d5431fa9a5ea05833ba8fd85f590b4db7e362)
- [Describe background job spans](https://github.com/appsignal/appsignal-ruby/commit/fb4f0396996116da384a687a85abe2ce57108c0f)
- [Mark render events as templating work](https://github.com/appsignal/appsignal-ruby/commit/cfdf2f4483df285917c460be5955a11c1e170b96)
- [Record the error type on spans that failed](https://github.com/appsignal/appsignal-ruby/commit/8059652c9e4333eace76b940ab8ec2d11e93efb8)

### [Fix the negated include_event matcher (#1564)](https://github.com/appsignal/appsignal-ruby/pull/1564)

- [Fix the negated include_event matcher](https://github.com/appsignal/appsignal-ruby/commit/618eb0623cad588595ed1d72643326fd54228fe6)

### [Add OpenTelemetry support for Excon integration (#1566)](https://github.com/appsignal/appsignal-ruby/pull/1566)

- [Test Excon suppression under Faraday again](https://github.com/appsignal/appsignal-ruby/commit/eb12d0236674da6da7c0d660af5c212eb4377fdd)
- [Mark Excon spans as client spans](https://github.com/appsignal/appsignal-ruby/commit/8ae58234ab20b33166b536f98033b8095ee432f3)
- [Inject trace context into Excon requests](https://github.com/appsignal/appsignal-ruby/commit/ea8fdfa72209b99247a514bfb37b3f0a48b7cb9f)
- [Describe Excon requests on their span](https://github.com/appsignal/appsignal-ruby/commit/5db9a4867ed2c6bed4c0cfdab4e5c24ec1264ba4)

### [Drop the duplicate queue start attribute](https://github.com/appsignal/appsignal-ruby/commit/7e4da7dcc9273f7f504f448aec4f50861eae642b)

The queue start event on the transaction's root span carried the queue
start time twice, as the event's timestamp and as an attribute of the
same name. A span event carries its own time, so the attribute said
nothing the event did not. Only the timestamp is set now, which is what
the breadcrumb event a little further up the file already does.

The specs asserted the attribute and nothing about the timestamp, even
where an example's own name said it was about the time the queue
started. They assert the timestamp now.

### [Report a signalled runner script as a failure](https://github.com/appsignal/appsignal-ruby/commit/51434c840389d3ce030a98acd8efc17675573423)

The integration runner checked a finished script's exit code to decide
whether it had failed. A process killed by a signal has no exit code, so
reading one raised a NoMethodError on nil.

That hid both what happened to the script and the output the check
prints to explain it. Ask the status whether it succeeded instead,
which answers for a signalled process too, and name the status in the
failure.

### [Express per-event treatment through event formatters (#1568)](https://github.com/appsignal/appsignal-ruby/pull/1568)

- [Remove an event formatter that never ran](https://github.com/appsignal/appsignal-ruby/commit/176186e4c0abff7a535317f38886b4e6a54a5ac2)
- [Let a formatter describe an event with no title](https://github.com/appsignal/appsignal-ruby/commit/36d2442c99134f7a803584421cb56fe80b565d88)
- [Let a formatter declare its span kind](https://github.com/appsignal/appsignal-ruby/commit/e16b525759b490b8c2b96656c69848d6e58e6ec0)
- [Let a formatter describe its own span](https://github.com/appsignal/appsignal-ruby/commit/bf0b2654778c25179d226dd7f3ad87d3298f2ac7)
- [Let a formatter defer to another integration](https://github.com/appsignal/appsignal-ruby/commit/efa753699c3063b12478ff09ec2b366fb3a7bf4b)
- [Let a formatter declare its instrumentation scope](https://github.com/appsignal/appsignal-ruby/commit/9ed4c09010df94348514dd7fa1d4577298011098)

### [Trim the unreleased changesets](https://github.com/appsignal/appsignal-ruby/commit/9076862677444c7772e6f8fbe1e3d5a646a4849d)

The changesets explained how each change works and why the old
behaviour was wrong. A changelog entry is read by someone deciding
whether it is about them, so it needs the symptom, the new behaviour
and anything they have to act on, not the mechanism. The collector mode
entry also said that no AppSignal data flows through the OpenTelemetry
SDK yet, which will not be true when this ships.

### [Match monitor's arguments in monitor_and_stop (#1573)](https://github.com/appsignal/appsignal-ruby/pull/1573)

- [Match monitor's arguments in monitor_and_stop](https://github.com/appsignal/appsignal-ruby/commit/b068e499edfe5c2430484d4626d51714e3fd6c3b)

### [Default instrument_sql to a client span (#1577)](https://github.com/appsignal/appsignal-ruby/pull/1577)

- [Default instrument_sql to a client span](https://github.com/appsignal/appsignal-ruby/commit/1e3cd2c68c0515b0893bcbbdae64772b10357d3e)

### [Suppress native events even when hooks are off (#1579)](https://github.com/appsignal/appsignal-ruby/pull/1579)

- [Suppress native events even when hooks are off](https://github.com/appsignal/appsignal-ruby/commit/3429061de74aa04f07ddf987f94c9001c7358811)

### [Name the real SQL engine in db.system.name (#1580)](https://github.com/appsignal/appsignal-ruby/pull/1580)

- [Stop the sentinel overwriting db.system.name](https://github.com/appsignal/appsignal-ruby/commit/807d01a18d7e171b6c03e5ac84e203b741c25482)
- [Apply formatter attributes to dry-monitor events](https://github.com/appsignal/appsignal-ruby/commit/eb8748de5ff4324b09746999a83c3a4a11b466c5)
- [Let record_event take opentelemetry_attributes](https://github.com/appsignal/appsignal-ruby/commit/2b036e6f94e85e5dbaf8e78e658f0b300d18d39f)
- [Name the engine for ActiveRecord, Sequel, ROM](https://github.com/appsignal/appsignal-ruby/commit/79171d6d0455ed4c01cb9af80795cd2077880bf2)
- [Name the real engine for DataMapper](https://github.com/appsignal/appsignal-ruby/commit/546bde273d816d15c6e934315f13ae9e79676b4d)


lib/appsignal/logger/opentelemetry_backend.rb
```diff
@@ -0,0 +1,78 @@
...
+      }.freeze
+
+      MUTEX = Mutex.new
+
+      class << self
+        def emit(group, severity, format, message, attributes)
+          number, text = OTEL_SEVERITY_MAP.fetch(severity, [0, nil])
+          otel_attributes = Appsignal::OpenTelemetry::Attributes.format(attributes)
+          otel_attributes["appsignal.group"] = group.to_s
+          otel_attributes["appsignal.format"] = FORMAT_NAMES.fetch(format, "autodetect")
```
unflxw, line 46: Support for these attributes was implemented in https://github.com/appsignal/appsignal-collector/pull/384 and https://github.com/appsignal/appsignal-processor-rs/pull/2138.

lib/appsignal/helpers/instrumentation.rb
```diff
@@ -715,7 +715,7 @@ def add_headers(headers = nil, &block)
       # Breadcrumbs can be used to trace what path a user has taken
       # before encountering an error.
       #
-      # Only the last 20 added breadcrumbs will be saved.
+      # At most 20 of the added breadcrumbs will be saved.
```
unflxw, line 902: This is a genuine divergence in behaviour between collector mode and agent mode: agent mode saves the last 20, while collector mode saves the first 20. We could choose to make them both converge on the collector mode's behaviour, but it would be a breaking change for the agent.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      # The OTel span name is what the collector surfaces as the event's
+      # label in the trace UI, so prefer the human-readable `title` (e.g.
+      # "User Load", "GET https://example.com") and fall back to the AS::N
+      # `name` (e.g. "sql.active_record") when no formatter supplied a title.
+      # The machine name still rides along in `appsignal.category` so it is
+      # not lost once the title wins the span name -- it keeps the event's
+      # grouping key available for later filtering.
+      def write_event_name_attributes(span, name, title)
+        span.name = title && !title.empty? ? title : name
+        span.set_attribute("appsignal.category", name)
```
unflxw: This does not have any specific purpose in our OpenTelemetry UI, but will be shown as an internal attribute.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      # the trace UI. The gem only has a single merged params blob, so route it
+      # by namespace: message/job (CONSUMER-kind) transactions use the
+      # function-parameters channel, everything else (web-style, SERVER-kind)
+      # uses the request-payload channel.
+      def params_attribute
+        if SPAN_KIND_BY_NAMESPACE.fetch(@namespace, DEFAULT_SPAN_KIND) == :consumer
+          "appsignal.function.parameters"
+        else
+          "appsignal.request.payload"
+        end
```
unflxw: This is a bit of a hack -- it makes the not-necessarily-true assumption that the "params" in transactions with consumer root spans (from background jobs being processed) will represent function parameters, and otherwise they will represent a request payload.

Since agent mode didn't really have this differentiation to begin with, I think it's fine. We could choose to expose explicit setters for each, but I think that's beyond the scope of this work at the moment.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+          tracer.start_span(name, :with_parent => opentelemetry_context, :kind => kind)
+        elsif remote && kind == :consumer
+          tracer.start_root_span(
+            name,
+            :kind => kind,
+            :links => [::OpenTelemetry::Trace::Link.new(remote)]
+          )
+        else
+          tracer.start_root_span(name, :kind => kind)
+        end
```
unflxw: Another sneaky assumption: when an OpenTelemetry trace context is passed when creating a transaction, when the transaction has a "server-like" namespace, then the root span of that transaction will be the child of that trace context's span, and when the transaction has a "consumer-like" namespace, then the root span of the transaction will have a span link to that trace context's spans. This matches the difference in behaviour between background job libraries and HTTP client/server libraries in the OpenTelemetry Ruby ecosystem.

At the Transaction API level, for custom instrumentation, we probably would want to expose an explicit flag to choose the kind of relationship. But generally the assumption with collector mode is that this is a compatibility layer, and newly implemented custom instrumentation would use the OpenTelemetry SDK directly, so this wasn't a priority.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+
+        namespace = display_namespace(@namespace)
+        Appsignal::Metrics::OpenTelemetryBackend.add_distribution_value(
+          "transaction_queue_duration", duration_ms, :namespace => namespace
+        )
+        Appsignal::Metrics::OpenTelemetryBackend.add_distribution_value(
+          "transaction_queue_duration", duration_ms,
+          :namespace => namespace, :hostname => hostname
+        )
+      end
```
unflxw, line 481: Our obscure queue start functionality measures the time that an HTTP request spent "queueing" in a reverse proxy before being handed out for the application server instance to handle. This recreates the metrics that the agent emits when this functionality is used.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      # wins). The flag must be written before the span finishes -- attributes
+      # set on an ended span are dropped. Tearing the span down here also
+      # detaches the context, so a discarded transaction can't leak its root
+      # span as the current OTel context onto the thread.
+      def discard
+        return if @completed
+
+        @span&.set_attribute("appsignal.ignore_subtrace", true)
+        teardown
+      end
```
unflxw, line 417: In agent mode, the agent discards transactions marked with the discard flag on completion. In collector mode, we piggy-back the existing `appsignal.ignore_subtrace` magic attribute for the same effect.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+        current_span.add_event(
+          "appsignal.breadcrumb",
+          :timestamp => Time.at(breadcrumb[:time]),
+          :attributes => {
+            "category" => breadcrumb[:category],
+            "action" => breadcrumb[:action],
+            "message" => breadcrumb[:message],
+            "metadata" => JSON.generate(breadcrumb[:metadata] || {})
+          }
+        )
```
unflxw, line 380: Explicit support for breadcrumbs in the OpenTelemetry UI was added here: https://github.com/appsignal/appsignal-server/pull/16442

It is less informative than looking at the corresponding events in the trace timeline, as the trace timeline will correlate them with the span in which they took place, but it preserves the UI expectation of existing breadcrumb users during the agent-to-collector transition. We could phase it out in the future, along with breadcrumbs at large.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      # (keys match the processor's `ErrorSubCause`); separate cause events would
+      # each become their own incident.
+      def set_error(class_name, message, backtrace, causes, _root_cause_missing)
+        span = current_span
+
+        attributes = {
+          "exception.type" => class_name,
+          "exception.message" => message.to_s,
+          "exception.stacktrace" => Array(backtrace).join("\n"),
+          "appsignal.alert_this_error" => true
```
unflxw, line 327: All errors added manually by `set_error`/`add_error` to a transaction are deemed alertable, matching the behaviour that agent mode implements using transaction duplicates. Unlike in transaction mode, the error is added to the span of the current event, not to the transaction's root span, and no duplicates of the transaction are made.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      end
+
+      def write_event_body_attributes(span, body, body_format)
+        return if body.nil? || body.empty?
+
+        if body_format == Appsignal::EventFormatter::SQL_BODY_FORMAT
+          span.set_attribute("db.query.text", body)
+          span.set_attribute("db.system.name", SQL_DB_SYSTEM)
+        else
+          span.set_attribute("appsignal.body", body)
```
unflxw, line 780: Support for this attribute in the OpenTelemetry UI was implemented here: https://github.com/appsignal/appsignal-server/pull/16430

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+        span.name = title && !title.empty? ? title : name
+        span.set_attribute("appsignal.category", name)
+      end
+
+      def write_event_body_attributes(span, body, body_format)
+        return if body.nil? || body.empty?
+
+        if body_format == Appsignal::EventFormatter::SQL_BODY_FORMAT
+          span.set_attribute("db.query.text", body)
+          span.set_attribute("db.system.name", SQL_DB_SYSTEM)
```
unflxw: The `db.system.name` attribute triggers SQL query sanitisation in the collector, while `db.query.text` is shown as "Query" by the OpenTelemetry UI.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+          attributes["appsignal.error_causes"] = JSON.generate(
+            causes.map do |cause|
+              {
+                "name" => cause[:name],
+                "message" => cause[:message],
+                "lines" => cause[:backtrace] || []
+              }
+            end
+          )
+        end
```
unflxw, line 344: The changes to support error causes being emitted in this way are implemented in this (not merged at the time of writing) processor PR: https://github.com/appsignal/appsignal-processor-rs/pull/2197

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      # graph). Like the agent, we record the delta and never shift span timing.
+      def set_queue_start(start)
+        return unless start && start > QUEUE_START_MIN
+
+        @queue_start = start
+        @span.add_event(
+          "appsignal.queue_start",
+          :timestamp => Time.at(start / 1000.0),
+          :attributes => { "appsignal.queue_start" => start }
+        )
```
unflxw, line 235: In addition to emitting the metrics that the agent would derive from the queue start, we also emit the queue start time itself as an event on the root span.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+      def tracer
+        ::OpenTelemetry.tracer_provider.tracer(TRACER_NAME, Appsignal::VERSION)
+      end
+
+      # The open event span, or the root span when no event is open. Not the OTel
+      # current span, which may belong to another instrumentation.
+      def current_span
+        span, _token = @event_stack.last
+        span || @span
+      end
```
unflxw, line 614: The "golden rule" of the AppSignal-to-OpenTelemetry interoperability story is that we don't write AppSignal-flavoured attributes and events to whatever is the current span in the active OpenTelemetry context, preventing AppSignal from "polluting" spans from external OpenTelemetry instrumentations. Instead, we keep a reference to the last AppSignal span created by the transaction: either the transaction's root span, or the span for the currently active event.

This way, whenever we add AppSignal-flavoured events or attributes, we always add them to one of our spans, either by using this helper or by referencing `@span` (the transaction's root span) directly -- we never add them to the current span in the active OpenTelemetry context.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+        # Transaction#initialize sets the namespace directly without calling
+        # set_namespace, so emit the attribute from here.
+        @span.set_attribute("appsignal.namespace", display_namespace(namespace)) if namespace
+      end
+
+      # `opentelemetry_kind` (e.g. `:client` for an outgoing HTTP request) is set
+      # at span creation because OTel span kind is immutable afterwards. `nil`
+      # leaves the SDK default (INTERNAL).
+      def start_event(opentelemetry_kind: nil)
+        span = tracer.start_span(EVENT_SPAN_PLACEHOLDER_NAME, :kind => opentelemetry_kind)
```
unflxw: Note that we don't pass an explicit `current_span` parent when creating the span for an event -- we allow the span an AppSignal event to be a child of whichever the current active span in the OpenTelemetry context is. This allows AppSignal events and OpenTelemetry spans to nest within each other.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Note that, when creating a transaction's span, it's either a root span, or it's a parent of the span from the remote trace context -- it's never a child of the current span in the active OpenTelemetry context. This ensures that transactions always translate to traces, and aren't accidentally nested as spans inside the traces created by an external OpenTelemetry instrumentation.

EDIT: this was actually not true at the time of writing, there was a bug that was fixed in https://github.com/appsignal/appsignal-ruby/pull/1535/commits/e65c4fd3a28f80191f492583c052bf2f84d80e3c

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,454 @@
...
+        end
+      end
+
+      def otel_header_name(env_key)
+        if env_key.start_with?("HTTP_")
+          env_key.delete_prefix("HTTP_").downcase.tr("_", "-")
+        elsif env_key.start_with?("CONTENT_")
+          env_key.downcase.tr("_", "-")
+        end
+      end
```
unflxw, line 718: This is a divergence between agent mode and collector mode, in that the "Environment" section in agent mode contains other Rack CGI-flavoured attributes in it as well as HTTP headers, but in collector mode, only actual HTTP headers will be emitted as `http.request.header.*` attributes, and all other CGI-flavoured attributes in it from Rack will be dropped.

spec/lib/appsignal/integrations/excon_spec.rb
```diff
@@ -0,0 +1,60 @@
+if DependencyHelper.excon_present?
+  require "excon"
+  require "appsignal/integrations/excon"
+  require "appsignal/integrations/excon/appsignal_middleware"
+
+  # Integration test against the real Excon gem (the hooks/excon_spec.rb suite
+  # stubs Excon). Verifies, end to end, that the inject-only middleware writes
+  # trace context onto a live outgoing request while the instrumentor's client
+  # event span is current -- the ordering the stubbed suite can't prove.
+  describe "Excon integration" do
```
unflxw, line 8: The existing Excon integration was previously untested -- this tests it just enough to assert that trace context propagation takes place.

unflxw (commented):
I've attempted to clarify and make explicit some decisions contained within this in the comments. Other decisions to point out, many of which were also present in #1515:

- The agent still runs. This provides host metrics, and preserves compatibility with out StatsD and NGINX metrics support. I believe this is a good idea and would like to backport it to Python's collector mode.
- The OpenTelemetry gems must be installed manually in order to keep compatibility with Ruby 2.7, for which no compatible version of the OpenTelemetry gems exists. This is a significant hurdle for the installation process. We can soften it with a meta-gem that installs those dependencies, so that at least it's still a one-liner. We can also bump the gem's Ruby version requirement to Ruby 3.1, which regardless of the gem's installation requirement, is a requirement for collector mode as currently implemented.
- The oddities and differences in how OpenTelemetry trace context propagation is done in the background job libraries mirror the oddities of the corresponding OpenTelemetry contrib instrumentations, or their closest equivalent. This attempts to provide wire compatibility between the two, such that you could enqueue a background job in a collector mode AppSignal-instrumented application, process it in an OpenTelemetry-instrumented application pointed at the same collector, and have the traces link correctly.

lib/appsignal/transaction.rb
```diff
@@ -734,23 +751,99 @@ def run_before_complete_hooks
...
+    # all block-set metadata merges onto the root span. (Per-error metadata
+    # isolation is deferred -- the processor/UI does not read per-event
+    # attributes yet.)
+    def run_error_blocks
+      @error_blocks.each_value do |blocks|
+        self.class.with_transaction(self) do
+          blocks.each { |block| block.call(self) }
+        end
+      end
+    end
```
unflxw, line 1112: I'm realising now that this is likely not the right behaviour. We probably want to run error blocks immediately in collector mode. Unlike in agent mode, where we need to store them in order to fabricate duplicates at completion time, I see no benefit to storing them and running them later in collector mode.

I don't think error blocks should be used like this (in collector mode, they likely should not be used at all) but if the block contains things that should be attached to the current span, such as breadcrumbs (🌵), other errors (😨) or custom OpenTelemetry instrumentation (💀) then, by running them immediately, those would be attached to the right point in the span tree.

(to be clear: none of those are things that an error block _should_ contain)

unflxw: Fixed in c8fbb047.

The one non-obvious part is `@error_blocks`. Collector mode no longer stores the block there, but it still registers the error as a key, because we rely on those keys for `last_errors`, the distinct-error limit, and the duplicate check. And since blocks now run when added, `report_errors` only has work to do in agent mode, so `run_error_blocks` is gone.

Added a collector-mode test to lock it in: an error whose block adds a breadcrumb, asserting the breadcrumb lands on the event span, not the root. Fails without the change.

lib/appsignal/config.rb
```diff
@@ -202,21 +214,29 @@ def dsl_config_file?
...
       :send_session_data => "APPSIGNAL_SEND_SESSION_DATA"
     }.freeze
 
     # @!visibility private
     ARRAY_OPTIONS = {
       :dns_servers => "APPSIGNAL_DNS_SERVERS",
+      :filter_attributes => "APPSIGNAL_FILTER_ATTRIBUTES",
+      :filter_function_parameters => "APPSIGNAL_FILTER_FUNCTION_PARAMETERS",
       :filter_metadata => "APPSIGNAL_FILTER_METADATA",
       :filter_parameters => "APPSIGNAL_FILTER_PARAMETERS",
```
tombruijn, line 249: I see you added warnings about if the agent options are used in collector mode.
Have you thought about how people upgrading would migrate their config? Do we have a hard requirement on people updating the config options? For example, if this config is set, it won't start/boot/ignore it. I'm asking in case you already have an idea for this but if not we can decide later.

We could also apply the value from this config option to the other two OpenTelemetry-specific ones as leading if set so it doesn't accidentally start reporting the previously filtered values on update.

spec/integration/stop_spec.rb
```diff
@@ -15,9 +15,6 @@
 
     output = runner.output
 
-    # Make sure the app exited properly
-    expect(runner.status.exitstatus).to eq(0)
```
tombruijn, line 19: Why don't we want to check this anymore?

lib/appsignal/helpers/instrumentation.rb
tombruijn: Why does it change to the first 20 breadcrumbs here?

Looks like it's because we immediately set the events on the span and we have a guard in front that only allows 20 to be set and exits early if it's higher than that, instead of keeping the breadcrumbs in memory until the span closes.

Could we get rid of the limitation or would that break things?

lib/appsignal/transaction/opentelemetry_backend.rb
tombruijn: We'll probably want to rename this to something like `queue_duration`, and not mention `transaction` as a context. We will get to this as we continue with the metrics migration.

lib/appsignal/transaction/opentelemetry_backend.rb
tombruijn: This is maybe a bit more work to do now, so let's pick it up later, but the way the OpenTelemetry record_exception helper works should hopefully provide us with all the information. 
If we go that route we also support errors with error causes reported directly from OpenTelemetry.

lib/appsignal/config.rb
unflxw: This follows how we approached it for Python. What we do is, if any config options that have no effect on agent mode are set, we warn about them, recommend an alternative config option if a close enough equivalent exists, and ignore them.

Note that it won't accidentally start reporting the previously filtered values on update, as collector mode needs to be explicitly enabled by setting the `collector_endpoint` environment variable. Otherwise, agent mode continues to honor the config options.

lib/appsignal/helpers/instrumentation.rb
unflxw: > Could we get rid of the limitation or would that break things?

This work at large is trying to mimic the existing transaction limitations as closely as possible. I do not know why the limitation was set to this specific number, or why it was set at all for agent mode, or how valuable it is in collector mode, if at all. Yes, we could probably get rid of this limitation entirely, at least in collector mode, but I'm trying to limit the surface of decisions that are re-evaluated.

The agent mode behaviour is to store the last 20 breadcrumbs for **the transaction**, then set them before the span is closed.

In collector mode, we set breadcrumbs on **the event's span** instead, which I would argue is an improvement, in that rather than the breadcrumbs constituting a "parallel timeline" as they do in agent mode, it correlates them to a specific point in the trace timeline.

This, however, means that, at the latest, we could wait until _the event's span_ closes to decide which breadcrumbs we send -- meaning we can't mimic the agent's behaviour of sending the last twenty _in the whole transaction_ without giving up assigning them to the right span. 

Since I thought the improvement was valuable, and we couldn't do _the last 20_, I thought it best to enforce the core of the limitation as _"at most 20"_. We _can_ keep a **per-transaction** (not per-span!) count of how many breadcrumb events we've emitted, and emit no more once we've already emitted twenty.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: The goal was to preserve compatibility with existing customer dashboards that could be using this metric. The name given to the metric is the name that the agent/processor pipeline (can't recall which one at the moment) gives to the metric that it derives from this value.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Yeah, would be nice. It didn't seem to me that `record_exception` explicitly supports it, but maybe it's implicit in the backtrace information it provides. This may be version-dependent. Could you please open an issue for looking into this in the processor repo?

spec/integration/stop_spec.rb
unflxw: Not sure, I'll look into it. Thanks for flagging.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: I think we may want to remove this, given the changes in https://github.com/appsignal/appsignal-ruby/pull/1551, where we're effectively using the group as the OpenTelemetry scope (in our default instrumentations -- this does not apply out of the box for custom instrumentation, as the existing `start_event`/`finish_event` API means the category is not known at span creation time)

lib/appsignal/transaction/opentelemetry_backend.rb
tombruijn: Note: remove the attribute. It's duplicate data.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,477 @@
...
+          :attributes => { "appsignal.queue_start" => start }
+        )
+      end
+
+      # Transaction metadata (request path, method, ...) has no dedicated OTel
+      # attribute, but it is the same shape as tags and the collector/trace UI
+      # already surface `appsignal.tag.*`, so emit metadata as a tag.
+      def set_metadata(key, value)
+        @span.set_attribute("appsignal.tag.#{key}", value)
+      end
```
lipskis, line 243: Do we need to put these under `appsignal.tag.*`? There's nothing special about the `appsignal.tag.*` attribute prefix. I'm thinking maybe it's better to use something like `appsignal.metadata.#{key}` so that we have the option of treating them differently in the UI if there's a need for it?

unflxw: I can't answer this too confidently, because honestly I've never understood the difference between the "metadata" and "tag" concepts in the transaction API, so I can't tell you for sure if being able to treat them differently would or wouldn't be valuable. 😅

What I do know is that, while they existed as different things at some point, as we've iterated on the processing pipeline, we seem to have collapsed both into tags. For newer integrations, OpenTelemetry-based or otherwise, we've not built a metadata equivalent.

As to why we put them in `appsignal.tag.*`, it's precisely because the UI does surface that today, that's the only thing that's special about it.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Removed. The event keeps its timestamp, which is where the queue start time was already being recorded, so nothing is lost by dropping the attribute.

This also matches what `add_breadcrumb` does a little further up. A span event carries its own time, so the time does not need to be repeated as an attribute.

spec/integration/stop_spec.rb
unflxw: We do still check it. The check moved out of this spec and into `Runner#run`, which now raises when a runner script does not exit successfully, and includes the script's output in the failure message.

That means the check covers every integration runner script now, not just this one. Leaving it here as well would have asserted the same thing twice.

Sorry for the confusion. The move happened in the same commit that removed this line, so nothing in the diff pointed at where it went.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: This is done. The `appsignal.category` attribute is gone, removed along with the rest of the changes that put the event category in the span name.

The span name now leads with the category, so the grouping key is still there to read. Several specs assert that the attribute is no longer set, to keep it from coming back.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Opened https://github.com/appsignal/appsignal-processor-rs/issues/2278 to look into this in the future. It is a placeholder, so please add anything I have described badly.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: This is no longer a guess. Params now arrive on named channels, so the transaction is told which kind it is being given rather than the backend working it out from the root span's kind.

There are explicit setters for each: `add_request_payload` and `add_function_parameters`. Done in [`ab7ee065`](https://github.com/appsignal/appsignal-ruby/commit/ab7ee06569c1f6507f62995b0cfca9d7a69979e3), as part of #1552.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: The explicit flag this asks for exists now. A transaction takes `opentelemetry_kind` and `opentelemetry_relationship`, so whether the root span parents to the incoming context or links to it is stated by the caller instead of being derived from the namespace.

Done in [`32f5b746`](https://github.com/appsignal/appsignal-ruby/commit/32f5b7468e070aeeb04fc900131d5dfde21303ae), as part of #1552.

spec/lib/appsignal/integrations/excon_spec.rb
unflxw: The Excon integration has proper coverage now, so this no longer only tests trace context propagation.

#1563 rebuilt the integration and tested it against the real Excon gem, rather than a stub that called the instrumentor by hand. That found three things the old integration got wrong, including Excon requests being reported as taking almost no time. The OpenTelemetry side then came back on top of the rebuilt integration, in the commits from #1566.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -0,0 +1,761 @@
...
+      # name as the title as well; in that case the name is not repeated.
+      def write_event_span_name(span, name, title)
+        has_title = title && !title.empty? && title != name
+        span.name = has_title ? "#{name} (#{title})" : name
+      end
+
+      def write_event_body_attributes(span, body, body_format)
+        has_body = !body.to_s.empty?
+
+        if body_format == Appsignal::EventFormatter::SQL_BODY_FORMAT
```
matsimitsu, line 770: It seems the `db.query.text` text is only populated when it's a SQL-ish body format, while the processor does it for anything that is a database. (e.g. MongoDB and Redis).

Do we want to do this for all database-like fields, or do we purely want to stick here to SQL?


| event | this backend | ingestion path |
|---|---|---|
| `query.mongodb` (`integrations/mongo_ruby_driver.rb:92`, `EventFormatter::DEFAULT`) | `appsignal.body` | `db.query.text` |
| `search.elasticsearch` (`event_formatter/elastic_search/search_formatter.rb`, two-element `format`) | `appsignal.body` | `db.query.text` |
| `query.redis` | `appsignal.body` | `db.query.text` |

lib/appsignal/opentelemetry.rb
```diff
@@ -0,0 +1,336 @@
...
+          "Error shutting down OpenTelemetry SDK: #{e.class}: #{e.message}"
+        )
+      end
+
+      # Build the OpenTelemetry Resource that carries AppSignal config to the
+      # collector. Attributes whose underlying option is nil or an empty array
+      # are omitted so the collector applies its own defaults.
+      def build_resource(config)
+        revision = config[:revision].to_s.empty? ? "unknown" : config[:revision]
+        service_name = config[:service_name].to_s.empty? ? "unknown" : config[:service_name]
```
matsimitsu, line 218: `service.name` falls back to the literal `"unknown"`, and `:service_name` is a new option defaulting to `nil` (`config.rb:154`), so an app that upgrades to collector mode without setting `APPSIGNAL_SERVICE_NAME` reports every span under `unknown`. The app's actual name goes to `appsignal.config.name` instead.

The processor then falls back to the app name. Do we maybe want to do the same here?

matsimitsu (commented):
Leaving two comments here for things that the processor does differently, which may be better? but also may not, because I don't have all the context for the choices made. Let me know what we want to do here :)

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Thanks for bringing this up! This is definitely intentional on our end -- using `db.query.text` here would make the collector attempt to apply SQL sanitisation to the MongoDB and Elasticsearch query within, potentially mangling it. The processor's legacy-to-OpenTelemetry pipeline does not have that problem, so it should be fine -- though note that an event's "body" is potentially anything, and there's no guarantee for it to be a query, or to mean that this event is a database event.

The event formatters for `query.mongodb` and `search.elasticsearch` (refactored into existence in #1568) do set the right value for the `db.system.name` attribute, which should [cause it to be colored and grouped as a DB span](https://github.com/appsignal/appsignal-server/blob/9a9447c989058e149c3e8f2c468262e263f24f0a/app/frontend/javascripts/ui/TraceTimeline/utils/spanGroups.js#L59) in the trace timeline. I also added a "Body" box for `appsignal.body`, similar in appearance to the existing "Query" box, in [appsignal-server#16430](https://github.com/appsignal/appsignal-server/pull/16430). 

That said, looking into this has flagged [some other behaviours](https://github.com/search?q=repo%3Aappsignal%2Fappsignal-server+db.query.text&type=code) that are specific to `db.query.text` throughout the server codebase, and I should look into whether they need to be extended for `appsignal.body`. Thank you!

lib/appsignal/opentelemetry.rb
unflxw: Thanks for bringing this up! I agree that the "unknown" fallback isn't ideal, but I'm not sure about falling back to the app name -- note that the app name can be any sort of string, which may look out of place as a service name.

I wonder if it would make sense to fall back to something neutral, like "unknown", but less you-did-something-wrong-sounding, like... "app"?

The [existing Python integration](https://github.com/appsignal/appsignal-python/blob/3e31f472906895012860ab6ce82988c3cdc78c9e/src/appsignal/opentelemetry.py#L293) also falls back to "unknown" in collector mode, so any change to one should also be done in the other. (This does not preclude the change, I'm just leaving a reminder for myself to keep it in sync)

Side note: this is also a good reminder to ensure that installation instructions for customers in collector mode ask them to set a service name.

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: > using db.query.text here would make the collector attempt to apply SQL sanitisation to the MongoDB and Elasticsearch query within, potentially mangling it

Nope, that's not true, I made it up! https://github.com/appsignal/appsignal-collector/blob/v0.11.2/collector/src/tracing/span.rs#L37 The collector checks that `db.system.name` says that it's an SQL system, so without `other_sql` (which _is_ set by the event's SQL format bit) it wouldn't attempt sanitisation.

So... hm, maybe we could always put them as `db.query.text`. But then we're potentially forcing DB coloring/treatment on custom instrumentation spans that use the body for any other purpose? I'm still not sure if that's a good idea.

PR #1544, opened 2026-07-14
Guard blocks handed to the transaction

unflxw:
Related to #1542 -- the issue this PR fixes (triggered by a bogus method call in a test setup, now removed in https://github.com/appsignal/test-setups/pull/380) is what allowed the OpenTelemetry context to leak in the first place. They each address one half of the issue: #1542 fixes accidentally making the current context the parent of what should always be a root span, this fixes a previous current context leaking into newer requests.

---

### [Guard blocks handed to the transaction](https://github.com/appsignal/appsignal-ruby/commit/21c7c8b810187943794c4e8403409aedc3e7fabb)

A transaction runs blocks supplied by user code in a few places: the
error block passed to `set_error`, `send_error`, and `report_error`, and
the `after_create` and `before_complete` hooks. Any of them can raise.

These blocks can run far from where they were defined. In agent mode an
error block runs at completion, not when the error is added. So a raise
propagated out of whatever drove creation or completion, surfacing the
error in unrelated code. In collector mode it also skipped the backend's
completion step, which left the transaction's OpenTelemetry context
attached to the fiber. That leaked context then became the parent of the
next request's root span on the reused thread, merging unrelated
requests into one trace. Agent mode never showed the leak because its
extension handle is not a per-fiber stack.

Run each of these blocks defensively. On failure, log the error and the
block's definition site, then carry on so creation and completion always
finish. The error is swallowed rather than re-raised, because the block
can run far from its caller and re-raising would surface it in code that
has nothing to do with the block.

### [Wrap transaction blocks at their entry points](https://github.com/appsignal/appsignal-ruby/commit/165ac380706fd9553a56e52175c7288ec074c24b)

The first version guarded each block where it runs. That wrapped the
error block in more than one place, including once from inside a block
passed to `internal_set_error`, which was redundant.

Wrap each error block once instead, where user code hands it in. It then
stays protected wherever it later runs. `send_error` and `report_error`
name themselves when they wrap the block, so the log points the customer
at the call they made, rather than at an internal method name.

The `after_create` and `before_complete` hooks keep their guard where
they run. They can be registered with a block or by pushing a method
onto the set, so there is no single entry point to wrap them at.

### DIFF
--- a/lib/appsignal/transaction.rb
+++ b/lib/appsignal/transaction.rb
@@ -627,7 +627,7 @@ module Appsignal
 
     # @!visibility private
     # @see Appsignal::Helpers::Instrumentation#report_error
-    def add_error(error, &block)
+    def add_error(error, source: nil, &block)
       unless error.is_a?(Exception)
         Appsignal.internal_logger.error "Appsignal::Transaction#add_error: Cannot add error. " \
           "The given value is not an exception: #{error.inspect}"
@@ -641,6 +641,11 @@ module Appsignal
         return
       end
 
+      # Wrap the block here, at the entry point, so it stays protected wherever
+      # it later runs: right away in collector mode, or at completion in agent
+      # mode. `source` names the helper the block was given to, when known, so
+      # the log points the customer at the right call.
+      block = protect(source ? "the block passed to #{source}" : "the error block", &block) if block
       internal_set_error(error, &block)
 
       # Mark errors and their causes as tracked so we don't report duplicates,
@@ -753,7 +758,7 @@ module Appsignal
         # completion.
         if block
           self.class.with_transaction(self) do
-            call_transaction_block(block, self, :source => "error")
+            block.call(self)
           end
         end
       else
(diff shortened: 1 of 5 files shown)


unflxw:
I've made this fix target `wip-otel-rework` (#1535) but it could be backported to `main`, as it is a fix for agent mode too.

lib/appsignal/transaction.rb
```diff
@@ -751,7 +751,11 @@ def internal_set_error(error, &block)
         # span -- breadcrumbs, nested errors, custom instrumentation -- then
         # lands where the error was reported, not on the root span at
         # completion.
-        self.class.with_transaction(self) { block.call(self) } if block
+        if block
+          self.class.with_transaction(self) do
+            call_transaction_block(block, self, :source => "error")
```
tombruijn: Would you want to differentiate between the three different handlers that now have the source set to 'error'? 

Like so for example:
```suggestion
            call_transaction_block(block, self, :source => "internal_set_error")
```

tombruijn (approved):

lib/appsignal/transaction.rb
unflxw: I'd like for it to be meaningful to the customer -- so `send_error`/`set_error`/`report_error`, but that information is lost at this point in the call tree. I could try and pass it down, or I could try to show all three names for readability.

As far as _we're_ concerned, in terms of our ability to debug it, we should get which one of `internal_set_error` or `report_errors_as_duplicates` it was that caused it from the backtrace itself, so that information is never lost. The method name in the error is for the _customer_, for them to be able to debug what it is they're doing in a block that caused it to error.

What does concern me, though, is that you mentioned it's called three different times. I think some of those are redundant with each other -- if `internal_set_error` calls its block with `call_transaction_block`, then `report_errors_as_duplicates` does not need to call its block with `call_transaction_block` _from inside a block passed to `internal_set_error`_. Thank you for flagging that.

unflxw: Following up on this: I went with wrapping at the entry point.

Each error block is now wrapped once, where user code hands it in, instead of at every place it later runs. That fixed the redundancy I mentioned. The duplicate-transaction path no longer wraps a block from inside a block passed to `internal_set_error`. It calls the already-wrapped block directly.

It also recovered the customer-facing name. `send_error` and `report_error` pass their own name when they wrap the block, so the log names the call the customer made rather than an internal method. `set_error` is left out on purpose, because its block runs inline at the call site, so a failure there already surfaces to the caller.

The `after_create` and `before_complete` hooks are the one exception. They can be registered either with a block or by pushing a method onto the set, so there is no single entry point to wrap them at. They keep their guard where they run.

PR #1551, opened 2026-07-20
Tag integration spans with instrumentation scope

unflxw:
The motivation for implementing this PR comes from the research into how our product exposes allocation counts that led to #1550, and specifically the sample breakdown. The OpenTelemetry-idiomatic equivalent of AppSignal category groups would be instrumentation scopes.

The changes in #1535 already emit the event category, from which we could extract the group, but if we were to build an equivalent to sample breakdowns or allocation breakdowns that was idiomatic to OpenTelemetry, we'd likely want to build it based on instrumentation scopes. Adding instrumentation scopes to each of our integrations enables that change, avoiding the need for retrofitting a category-based solution in the future to account for data from previous versions of the gem.

---

In collector mode, this makes each AppSignal integration emit its spans under its own OpenTelemetry instrumentation scope, named after the integration and tagged with the gem version. Grouping spans by scope lets the UI show per-scope statistics later, and it works the same way for spans sent by any other OpenTelemetry setup.

The commits come in implementation and test pairs, with the plain scope-passing folded together and the two changes that do more than pass a scope through, the ActiveSupport::Notifications bridge and the Rack layer, split out on their own.

---

### [Resolve OTel logger per emit, drop the cache](https://github.com/appsignal/appsignal-ruby/commit/06bb4e81529099fb08867feb6c193e4ac67c286a)

The OpenTelemetry logs SDK already caches the logger in its provider,
keyed by name and version. Caching the logger again in this backend
duplicated that, and the only thing the local cache added was avoiding a
provider registry lookup on each emit.

Resolve the logger from the provider on every emit instead. This is a
registry lookup, not a rebuild, and it always reflects the currently
configured provider. That removes the need for the test-only reset!
hook, which existed only to drop this cache when the provider was
swapped between tests.

Update the specs to match. Drop the "logger caching" examples and prove
instead that a swapped logger provider is picked up on the next emit
without any reset. Remove the reset! calls from the backend spec and
from the collector-mode shared context.

### [Add opentelemetry_scope to instrumentation](https://github.com/appsignal/appsignal-ruby/commit/fcae93fb98cc68070ecb3c4b0743d91ad43aca10)

Add an optional `opentelemetry_scope:` argument, a `[name, version]`
pair that names the OpenTelemetry instrumentation scope a span belongs
to. It is threaded from the instrumentation helpers and from
`Transaction.create` through the transaction to the backends, the same
way `opentelemetry_kind` already is. The instrument, instrument_sql,
monitor, monitor_and_stop, send_error and report_error helpers all take
it and forward it to the transaction they create.

In collector mode the OpenTelemetry backend resolves a tracer for that
scope and uses it for the root span and for each event span. A nil
scope, or one with a blank name, falls back to the default
`appsignal-ruby` scope at the gem version, so every span always carries
a scope. The agent backend accepts the argument and ignores it, because
agent mode has no notion of instrumentation scope.

This wires the mechanism only. Integrations start passing their own
scope in later changes. Until then every span keeps the default scope,
so behaviour is unchanged.

Add a `scope_of` helper to the collector-mode shared context so specs
can read a span's instrumentation scope, and cover the plumbing. The
scope is forwarded to the backend, the root and event spans are recorded
under the given scope, a blank or missing scope falls back to the
default, and the error and monitor helpers record their created
transaction under a passed scope.

### [Derive AS::N span scope from the event group](https://github.com/appsignal/appsignal-ruby/commit/1bfbc2491ce3d2b071e52fcdf4e133bb49f2e560)

ActiveSupport::Notifications bridges many Rails components through one
code path, so a single scope would lump them all together. Derive the
instrumentation scope from the event name group, which is the part after
the last dot. That puts each component under its own scope, such as
appsignal-ruby/active_record and appsignal-ruby/action_view. An event
name without a group falls back to the default scope.

### [Test AS::N derived instrumentation scope](https://github.com/appsignal/appsignal-ruby/commit/6f3004eb0c5a58e10a4e8a873906634649fbb5e3)

Assert the group-derived scope in collector mode: sql.active_record
under appsignal-ruby/active_record, sql.sequel under
appsignal-ruby/sequel, a dotted custom name under its group, and a name
with no group under the default appsignal-ruby scope.

### [Tag Rack and web framework spans with scope](https://github.com/appsignal/appsignal-ruby/commit/c7ade961150e08feb38f6d294abcbf6866e17238)

Give the Rack middleware an opentelemetry_scope option and pass it when
it creates the request transaction and its wrapping event. The generic
middleware, the response body wrapper and the Rack event handler all use
the appsignal-ruby/rack scope, so their spans are grouped under the Rack
scope in collector mode.

Each web framework middleware sets its own scope through this option:
appsignal-ruby/sinatra, appsignal-ruby/grape, appsignal-ruby/hanami,
appsignal-ruby/padrino and appsignal-ruby/rails. Their request spans are
grouped under the matching scope.

### [Test Rack and web framework span scope](https://github.com/appsignal/appsignal-ruby/commit/f83bf78fe1327e51ab0d5608435f91321314e649)

Assert in collector mode that the generic Rack middleware and the Rack
event handler record their request spans under the appsignal-ruby/rack
scope, and that the Sinatra, Grape, Hanami and Rails middlewares record
theirs under their framework scope. The Rails case covers the middleware
creating the transaction itself, since a full Rails stack usually has
the Rack event handler create it first under the Rack scope.

Update the Padrino loader spec too. The loader now registers the Sinatra
base instrumentation middleware with an opentelemetry_scope option, so
the middleware registration assertion expects it.

### [Tag remaining integration spans with scope](https://github.com/appsignal/appsignal-ruby/commit/0cb1e434c9f3af70fcb2287510148362f6281f5f)

Each of these integrations now passes its own OpenTelemetry
instrumentation scope, a [name, version] pair at the gem version, when
it creates a transaction or records an event. In collector mode their
spans are grouped under that scope.

Background jobs use appsignal-ruby/sidekiq, -que, -shoryuken, -resque,
-delayed_job and -active_job. HTTP clients use appsignal-ruby/net_http,
-http_rb, -faraday and -excon. Datastores use appsignal-ruby/redis,
-redis_client, -sequel, -mongo, -data_mapper and -dry_monitor. Action
Cable, Webmachine and Rake use appsignal-ruby/action_cable, -webmachine
and -rake.

Error reporting through report_error passes a scope too. Puma low-level
errors use appsignal-ruby/puma, the Rails error reporter uses
appsignal-ruby/rails and the at-exit hook uses appsignal-ruby/at_exit.
These only apply the scope when they create the transaction. When an
error is reported onto an existing transaction, that transaction keeps
its own scope.

### [Test remaining integration scopes](https://github.com/appsignal/appsignal-ruby/commit/6bc76a1f71990adf9a52e2f4f1b557686eaad6e7)

Assert in collector mode that each integration records its spans under
its own scope. This covers the background jobs, the HTTP clients and the
datastores, Action Cable, Webmachine and Rake, and the Puma, Rails and
at-exit error reporters. The Puma case also covers that reporting onto
an existing transaction keeps that transaction's scope rather than the
Puma one.

### [Add changeset for per-integration scopes](https://github.com/appsignal/appsignal-ruby/commit/ee4fd8ae71741336092a514223fd80d56886cf1b)

The instrumentation scope name is part of the data that leaves the
gem, so a customer can see it and group their spans by it. That makes
it a customer-visible change and it needs its own changelog entry,
separate from the entry for collector mode itself.

### DIFF
--- a/lib/appsignal/hooks/action_cable.rb
+++ b/lib/appsignal/hooks/action_cable.rb
@@ -37,10 +37,16 @@ module Appsignal
           env[Appsignal::Hooks::ActionCableHook::REQUEST_ID] ||= request_id
 
           transaction =
-            Appsignal::Transaction.create(Appsignal::Transaction::ACTION_CABLE)
+            Appsignal::Transaction.create(
+              Appsignal::Transaction::ACTION_CABLE,
+              :opentelemetry_scope => ["appsignal-ruby/action_cable", Appsignal::VERSION]
+            )
 
           begin
-            Appsignal.instrument "subscribed.action_cable" do
+            Appsignal.instrument(
+              "subscribed.action_cable",
+              :opentelemetry_scope => ["appsignal-ruby/action_cable", Appsignal::VERSION]
+            ) do
               inner.call
             end
           rescue Exception => exception
@@ -73,10 +79,16 @@ module Appsignal
           env[Appsignal::Hooks::ActionCableHook::REQUEST_ID] ||= request_id
 
           transaction =
-            Appsignal::Transaction.create(Appsignal::Transaction::ACTION_CABLE)
+            Appsignal::Transaction.create(
+              Appsignal::Transaction::ACTION_CABLE,
+              :opentelemetry_scope => ["appsignal-ruby/action_cable", Appsignal::VERSION]
+            )
 
           begin
-            Appsignal.instrument "unsubscribed.action_cable" do
+            Appsignal.instrument(
+              "unsubscribed.action_cable",
+              :opentelemetry_scope => ["appsignal-ruby/action_cable", Appsignal::VERSION]
+            ) do
               inner.call
             end
           rescue Exception => exception
--- a/lib/appsignal/integrations/active_support_notifications.rb
+++ b/lib/appsignal/integrations/active_support_notifications.rb
@@ -7,17 +7,30 @@ module Appsignal
       class << self
         BANG = "!"
 
-        # Events a dedicated AppSignal integration already records, so the
-        # generic notifications path must not record them a second time. The
-        # ActiveJob hook owns `enqueue.active_job` (it wraps the enqueue in its
-        # own event, with Rails' native notification nested inside), and the
-        # Faraday integration owns `request.faraday`.
+        # ActiveSupport::Notifications events whose span represents an outgoing
+        # call to a datastore, so they carry CLIENT kind in collector mode (to
+        # match the dedicated DB integrations). Kept deliberately narrow:
+        # `start_event` runs for every instrumented Rails event and span kind is
+        # immutable, so only genuine client calls belong here. Object
+        # instantiation (`instantiation.active_record`) is not a client call.
+        CLIENT_EVENT_NAMES = ["sql.active_record"].freeze
+
+        # Events a dedicated AppSignal integration already records with richer
+        # semantics, so the generic notifications path must not record them a
+        # second time. The ActiveJob hook owns `enqueue.active_job`: it wraps the
+        # enqueue in a producer event that also injects trace context, and the
+        # native notification fires nested inside it. The Faraday integration owns
+        # `request.faraday`: its middleware records the request as a client event
+        # and injects trace context, and Faraday's own instrumentation
+        # notification, if the user added that middleware, fires nested inside it.
         SUPPRESSED_EVENT_NAMES = ["enqueue.active_job", "request.faraday"].freeze
 
         def start_event(name)
           return unless record_event?(name)
 
-          Appsignal::Transaction.current.start_event
+          Appsignal::Transaction.current.start_event(
+            :opentelemetry_kind => CLIENT_EVENT_NAMES.include?(name.to_s) ? :client : nil
+          )
         end
 
         def finish_event(name, payload = {})
--- a/lib/appsignal/integrations/active_support_notifications.rb
+++ b/lib/appsignal/integrations/active_support_notifications.rb
@@ -13,7 +13,13 @@ module Appsignal
         # `start_event` runs for every instrumented Rails event and span kind is
         # immutable, so only genuine client calls belong here. Object
         # instantiation (`instantiation.active_record`) is not a client call.
-        CLIENT_EVENT_NAMES = ["sql.active_record"].freeze
+        #
+        # `sql.sequel` is emitted by the sequel-rails gem through
+        # ActiveSupport::Notifications, so it reaches us here rather than through
+        # the dedicated Sequel hook (which already tags its own query events as
+        # CLIENT). Including it keeps a Sequel query CLIENT regardless of which
+        # path records it.
+        CLIENT_EVENT_NAMES = ["sql.active_record", "sql.sequel"].freeze
 
         # Events a dedicated AppSignal integration already records with richer
         # semantics, so the generic notifications path must not record them a
--- a/lib/appsignal/integrations/active_support_notifications.rb
+++ b/lib/appsignal/integrations/active_support_notifications.rb
@@ -35,8 +35,26 @@ module Appsignal
           return unless record_event?(name)
 
           Appsignal::Transaction.current.start_event(
-            :opentelemetry_kind => CLIENT_EVENT_NAMES.include?(name.to_s) ? :client : nil
+            :opentelemetry_kind => CLIENT_EVENT_NAMES.include?(name.to_s) ? :client : nil,
+            :opentelemetry_scope => scope_for(name)
           )
+        end
+
+        # ActiveSupport::Notifications bridges many Rails components through this
+        # one path (`sql.active_record`, `render_template.action_view`, ...), so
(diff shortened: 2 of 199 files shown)


lib/appsignal/logger/opentelemetry_backend.rb
```diff
@@ -54,23 +52,13 @@ def emit(group, severity, format, message, attributes)
...
-        # mutex on the hot path, take the lock and re-check only on the
-        # first call.
+        # Resolve the OTel logger on each emit. The logger provider caches it
+        # by name, so this is a registry lookup rather than a rebuild, and it
+        # always reflects the currently configured provider.
         def logger
-          @logger || MUTEX.synchronize do
-            @logger ||= ::OpenTelemetry.logger_provider.logger(:name => "appsignal-logger")
-          end
+          ::OpenTelemetry.logger_provider.logger(:name => "appsignal-logger")
```
unflxw: This change stems from the realisation, during the research leading to the broader PR, that the `.logger` and `.tracer` methods in the logger and tracer providers respectively were ultimately doing the same memoization that our code was doing.

Rather than add a redundant layer of memoization and complicate our code by providing hooks to reset it in our tests, let the logger provider take care of it.

lib/appsignal/integrations/active_support_notifications.rb
```diff
@@ -35,10 +35,28 @@ def start_event(name)
...
+        # rather than lumping them under one. A name without a group falls back
+        # to the default scope in the backend.
+        def scope_for(name)
+          # Only names with a group (a dot) map to a component scope. A name
+          # without one has no component to attribute it to, so it falls back to
+          # the default scope in the backend (returning nil here).
+          parts = name.to_s.split(".")
+          return if parts.length < 2 || parts.last.empty?
+
+          ["appsignal-ruby-#{parts.last}", Appsignal::VERSION]
```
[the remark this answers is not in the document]
unflxw: Leaving this as `split(".")`. This path is not hot enough to be worth the readability hit of the `rindex` version.

lib/appsignal/hooks/action_cable.rb
```diff
@@ -37,10 +37,16 @@ def install_subscribe_callback
           env[Appsignal::Hooks::ActionCableHook::REQUEST_ID] ||= request_id
 
           transaction =
-            Appsignal::Transaction.create(Appsignal::Transaction::ACTION_CABLE)
+            Appsignal::Transaction.create(
+              Appsignal::Transaction::ACTION_CABLE,
+              :opentelemetry_scope => ["appsignal-ruby-action_cable", Appsignal::VERSION]
```
tombruijn: Is there any kind of naming convention/recommendation around instrumentation scope names? Like, could we use a `/` to 'namespace' the instrumentation scope names? So we could more easily recognize them / less likely conflict on other scope names.

```suggestion
              :opentelemetry_scope => ["appsignal-ruby/action_cable", Appsignal::VERSION]
```

tombruijn (approved):

lib/appsignal/hooks/action_cable.rb
unflxw: Convention varies per language and the specs don't go into language-specific detail. Choice quotes:

> [the tracer name must] uniquely identify the instrumentation scope, such as the instrumentation library (e.g. `io.opentelemetry.contrib.mongodb`), package, module or class name.

> [the instrumentation scope attributes should uniquely identify the emitting unit;] a typical approach to ensure uniqueness is to use the fully qualified name of the emitting software unit.

In Ruby, it is module names -- so `OpenTelemetry::Instrumentation::Sinatra` for example. I did consider that, but in our case, our instrumentations don't match exactly to module names -- the worst offender being our `ActiveSupport::Notification` instrumentation, which is a sort of meta-instrumentation that would need to either be told what (potentially non-existing) modules to attribute it too, or make them up on the fly. This syntax choice allowed us to "sneak in" our existing category groups, which fits in with the whole "seamless transition" goal.

Your slash proposal does seem like an improvement in that it makes the separation point more obvious, I'll implement that.

unflxw:
Merged into `wip-otel-rework` and closing, so the work is tracked in #1535 from here.

The commits were replayed onto the branch rather than merged through this pull request. Because the branch history was then rewritten, the commits on the branch have different SHAs than the ones here, which is why GitHub does not mark this as merged.

The eight code and test commits are on the branch. The ninth commit, which only added a changeset, was dropped: the collector mode work stays under the single changeset on `wip-otel-rework`, which will be reworked once the whole feature is done.

Where this met #1552, the keyword arguments were combined rather than either side taking precedence. `Transaction.create`, `Transaction#initialize` and both transaction backends now accept `opentelemetry_context`, `opentelemetry_scope`, `opentelemetry_kind` and `opentelemetry_relationship`, and so do `Appsignal.monitor`, `Appsignal.send_error` and `Appsignal.report_error`. A separate commit then documents `opentelemetry_scope`, so it generates as a `[String, String]` pair in the type signatures instead of `untyped`.

PR #1552, opened 2026-07-22
Set transaction OpenTelemetry kind and relationship explicitly

unflxw:
### [Set transaction span kind and relationship](https://github.com/appsignal/appsignal-ruby/commit/30fb99a8)

In collector mode the transaction backend derived the OpenTelemetry span
kind from the namespace, and derived how an incoming trace context was
used, parent or link, from that kind. Make both an explicit choice
passed to the transaction instead.

`Transaction.create` and the `monitor`, `send_error` and `report_error`
helpers gain `opentelemetry_kind` and `opentelemetry_relationship`
arguments. The kind is one of `:server`, `:consumer`, `:producer` or
`:internal` and defaults to `:server`. The relationship is one of
`:parent`, `:link`, `:both` or `:none` and defaults to `:parent`. The
helpers also accept an `opentelemetry_context`, so the relationship has
a context to act on.

Every integration now passes these attributes so its behavior does not
change. Background job integrations pass `:consumer` and `:link`. Web
integrations keep the server and parent defaults.

### [Split params into request and function channels](https://github.com/appsignal/appsignal-ruby/commit/879e0635)

The AppSignal collector keeps request parameters and function arguments
as separate attributes, but the gem only had one params blob and routed
it to one attribute or the other based on the namespace. Make the
channel an explicit choice instead.

Add `add_request_payload` for the parameters of an incoming request,
`add_function_parameters` for the arguments a job or function was called
with, and `add_query_parameters` for the parameters parsed from a
request's query string. Each has an `_if_nil` variant and a matching
`Appsignal.*` helper. This mirrors `add_params`. The existing
`add_params` and `set_params` still work and map to the request payload,
but warn once in collector mode to point at the explicit methods.

Each transaction backend declares a `params_mapping` from the logical
channels to storage buckets. The extension backend maps them all to one
`params` bucket, so agent mode keeps a single merged blob and behaves as
before. The OpenTelemetry backend maps the request payload, function
parameters and query parameters to separate buckets, so collector mode
keeps them apart as separate attributes. The transaction allocates one
params `SampleData` per bucket and routes every setter, reader and
`_if_nil` guard through the mapping, so the routing follows the backend
rather than any mode flag.

`set_empty_params!` empties every bucket, not only the one the legacy
params channel maps to. This is a deliberate choice: reporting no params
suppresses the function parameters too, rather than leaving a background
job's arguments to be added afterward, and it matches agent mode where
all channels share one bucket.

Every integration now picks the channel that matches its data. Web
integrations use the request payload, background job integrations use
the function parameters, and the Webmachine integration reports its
request query string as query parameters.

### [Parent background job spans to the enqueuer](https://github.com/appsignal/appsignal-ruby/commit/2a5ef7e3)

In collector mode a background job used to start its own trace and link
back to the span that enqueued it. Change the relationship to `:both` so
the job continues the enqueuer's trace as a child span and keeps the
link back to the enqueuing span.

This applies to Sidekiq, Resque, Que, Shoryuken and Active Job. Delayed
Job has no carrier for the trace context across the enqueue and perform
boundary. Its jobs stay their own trace, so the change is a no-op for
it.

[skip changeset]

### DIFF
--- a/lib/appsignal/helpers/instrumentation.rb
+++ b/lib/appsignal/helpers/instrumentation.rb
@@ -102,6 +102,14 @@ module Appsignal
       #   within the block with {#set_action}.
       #   This will not update the active transaction's action if
       #   {.monitor} is called when another transaction is already active.
+      # @param opentelemetry_kind [Symbol] In collector mode, the OpenTelemetry
+      #   span kind: one of `:server`, `:consumer`, `:producer` or `:internal`.
+      #   Defaults to `:server`.
+      # @param opentelemetry_relationship [Symbol] In collector mode, how an
+      #   incoming `opentelemetry_context` relates to this transaction's span:
+      #   one of `:parent`, `:link`, `:both` or `:none`. Defaults to `:parent`.
+      # @param opentelemetry_context In collector mode, an incoming OpenTelemetry
+      #   trace context to relate this transaction's span to.
       # @yield [] The block to monitor.
       # @yieldreturn [Object] The return value of the block
       # @raise [Exception] Any exception that occurs within the given block is
@@ -112,7 +120,13 @@ module Appsignal
       #
       # @see https://docs.appsignal.com/ruby/instrumentation/background-jobs.html
       #   Monitor guide
-      def monitor(action:, namespace: nil)
+      def monitor(
+        action:,
+        namespace: nil,
+        opentelemetry_context: nil,
+        opentelemetry_kind: nil,
+        opentelemetry_relationship: nil
+      )
         return yield unless Appsignal.active?
 
         has_parent_transaction = Appsignal::Transaction.current?
@@ -133,7 +147,12 @@ module Appsignal
           if has_parent_transaction
             Appsignal::Transaction.current
           else
-            Appsignal::Transaction.create(namespace || Appsignal::Transaction::HTTP_REQUEST)
+            Appsignal::Transaction.create(
+              namespace || Appsignal::Transaction::HTTP_REQUEST,
+              :opentelemetry_context => opentelemetry_context,
+              :opentelemetry_kind => opentelemetry_kind,
+              :opentelemetry_relationship => opentelemetry_relationship
+            )
           end
 
         begin
@@ -218,6 +237,14 @@ module Appsignal
       #
       # @since 0.6.0
       # @param error [Exception] The error to send to AppSignal.
+      # @param opentelemetry_kind [Symbol] In collector mode, the OpenTelemetry
+      #   span kind: one of `:server`, `:consumer`, `:producer` or `:internal`.
+      #   Defaults to `:server`.
+      # @param opentelemetry_relationship [Symbol] In collector mode, how an
+      #   incoming `opentelemetry_context` relates to this transaction's span:
+      #   one of `:parent`, `:link`, `:both` or `:none`. Defaults to `:parent`.
+      # @param opentelemetry_context In collector mode, an incoming OpenTelemetry
+      #   trace context to relate this transaction's span to.
       # @yield [transaction] yields block to allow modification of the
       #   transaction before it's send.
       # @yieldparam transaction [Transaction] yields the AppSignal transaction
@@ -226,7 +253,13 @@ module Appsignal
       #
       # @see https://docs.appsignal.com/ruby/instrumentation/exception-handling.html
       #   Exception handling guide
-      def send_error(error, &block)
+      def send_error(
+        error,
+        opentelemetry_context: nil,
+        opentelemetry_kind: nil,
+        opentelemetry_relationship: nil,
+        &block
+      )
         return unless Appsignal.active?
 
         unless error.is_a?(Exception)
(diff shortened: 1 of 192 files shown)


lib/appsignal/transaction/extension_backend.rb
```diff
@@ -60,6 +72,17 @@ def set_metadata(key, value)
...
 
+      # The agent has a single params slot, so every params channel maps to one
+      # `:params` bucket. The transaction merges the channels into it, and only
+      # the `:params` key ever reaches `set_sample_data`.
+      def params_mapping
+        {
+          :params => :params,
+          :request_payload => :params,
+          :function_parameters => :params
+        }
```
tombruijn: Can we move this to constant?

lib/appsignal/transaction.rb
```diff
@@ -387,10 +462,86 @@ def set_empty_params!
...
+    # @yield This block is called when the transaction is sampled. The block's
+    #   return value will become the new parameters.
+    # @yieldreturn [Hash<String, Object>, Array<Object>]
+    # @return [void]
+    #
+    # @see #add_function_parameters
+    def add_request_payload(given_params = nil, &block)
+      params_data(:request_payload).add(given_params, &block)
+    end
+    alias set_request_payload add_request_payload
```
tombruijn:
```suggestion
```

What if we leave this alias out for the new methods? We only have the `set_` method name versions for the old sample data because we don't want to break backward compatibility, but we don't need to add it for new methods.

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -152,16 +203,32 @@ def set_metadata(key, value)
...
+      # The collector keeps the request payload and the function parameters as
+      # separate attributes, so each gets its own bucket. Legacy `params` has no
+      # channel of its own, so it maps to the request payload (the web/server
+      # default), matching how the span kind defaults to `:server`.
+      def params_mapping
+        {
+          :params => :request_payload,
+          :request_payload => :request_payload,
+          :function_parameters => :function_parameters
+        }
```
tombruijn: Can we move this to a constant?

lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -152,16 +203,32 @@ def set_metadata(key, value)
...
+      # categories (session, custom data) are serialized as JSON; `environment`
+      # becomes request-header attributes; tags fan out to `appsignal.tag.*`.
+      # Unknown keys pass through as `appsignal.<key>` JSON so nothing is lost.
+      # Breadcrumbs never reach here (the backend emits them as span events);
+      # causes ride on the exception event (see #set_error).
       def set_sample_data(key, data)
         case key
-        when "params"
-          @span.set_attribute(params_attribute, JSON.generate(data))
+        when "request_payload"
```
tombruijn, line 233: Let's also add to the [query programs](https://docs.appsignal.com/opentelemetry/custom-instrumentation/attributes#request-query-parameters) case that we discussed in the call.

lib/appsignal/helpers/instrumentation.rb
```diff
@@ -602,16 +661,71 @@ def add_params(params = nil, &block)
...
+      #
+      # @see #add_function_parameters
+      def add_request_payload(params = nil, &block)
+        return unless Appsignal.active?
+        return unless Appsignal::Transaction.current?
+
+        transaction = Appsignal::Transaction.current
+        transaction.add_request_payload(params, &block)
+      end
+      alias set_request_payload add_request_payload
```
tombruijn: Same here about the aliases for these helpers. We probably don't need to add them.

tombruijn (approved):

lib/appsignal/transaction/opentelemetry_backend.rb
unflxw: Added. The Webmachine integration is rewired to use it too, which is worth flagging. Webmachine was the one integration collecting the request's query string specifically (`request.query`), and it was sending that as the request payload. It now sends it as query parameters. So in collector mode that data moves from `appsignal.request.payload` to `appsignal.request.query_parameters`.

unflxw:
Merged into `wip-otel-rework` and closing, so the work is tracked in #1535 from here.

The commits were replayed onto the branch rather than merged through this pull request. Because the branch history was then rewritten, the commits on the branch have different SHAs than the ones here, which is why GitHub does not mark this as merged.

The three commits are on the branch, ending at `Parent background job spans to the enqueuer`.

Where this met #1551, the keyword arguments were combined rather than either side taking precedence, so every signature accepts `opentelemetry_scope` alongside `opentelemetry_kind` and `opentelemetry_relationship`. Both backends list all four explicitly, keeping the property that an unexpected keyword raises in either mode. Each integration's `Transaction.create` call passes its instrumentation scope as well as its kind and relationship.

The specs that assert the exact keyword arguments reaching the backend now include `opentelemetry_scope` too, since it is always passed.

PR #1559, opened 2026-07-28
Trim shared frames from error cause backtraces

unflxw:
See also https://github.com/appsignal/appsignal-processor-rs/pull/2236 for the change to the processor that does _something_ with the omitted lines attribute.

### [Trim shared frames from error cause backtraces](https://github.com/appsignal/appsignal-ruby/commit/6019d2a3f15fead941c84e8c780f8d9e5b2baaa4)

In collector mode, an error's causes are sent as one JSON span-event
attribute, `appsignal.error_causes`, and each cause carried its full
backtrace.

A cause is raised inside the frames that led to the reported error, so
its backtrace ends with the same lines as the reported error's own
backtrace. For a web request that shared tail is the whole framework
and web server stack, and it was repeated for every cause. In a
measured three-deep cause chain from a Rails controller action, 88 of
each cause's 92 lines were a verbatim repeat, and the attribute came
to 21,080 characters.

The AppSignal Collector caps a span-event attribute value at 20,000
characters. When it truncates, the JSON is no longer valid, so the
collector could not read the causes and dropped the whole chain.
Nothing was shown under the error's causes in the trace.

Drop from each cause's backtrace the trailing lines it shares with the
reported error's backtrace. Those lines are already sent once, in
`exception.stacktrace`. What is left is where the cause was raised,
which is the part worth showing. On the measured payload this brings
the attribute down to about 904 characters. If a cause shares every
line, its first line is kept, because a cause with no lines leaves
nothing to show.

### [Report how many cause lines were omitted](https://github.com/appsignal/appsignal-ruby/commit/2fac3f6f6160bad46ad308d29116a1d5917e2496)

Each error cause in `appsignal.error_causes` no longer carries the
trailing backtrace lines it shares with the reported error's own
backtrace. A consumer of the attribute could not tell that lines were
left out, or how many of them there were.

Add a `lines_omitted` key to each cause, holding the number of trailing
lines that were dropped for that cause. The AppSignal Processor uses
the count to add a line reading "[88 repeated lines omitted]" to the
end of the cause's backtrace, so the trace shows how much is missing.

The count is the number of lines actually removed. When a cause shares
every line with the reported error, its first line is kept, so that
kept line does not count as removed. The key is left out entirely when
nothing was dropped, because a consumer reads a missing key as zero.

### DIFF
--- a/lib/appsignal/transaction/opentelemetry_backend.rb
+++ b/lib/appsignal/transaction/opentelemetry_backend.rb
@@ -283,7 +283,7 @@ module Appsignal
       # (keys match the processor's `ErrorSubCause`); separate cause events would
       # each become their own incident. Each cause only carries the part of its
       # backtrace that the reported error's backtrace does not already end with
-      # (see `lines_without_shared_tail`).
+      # (see `trim_shared_tail`).
       def set_error(class_name, message, backtrace, causes, _root_cause_missing)
         span = current_span
         error_lines = Array(backtrace)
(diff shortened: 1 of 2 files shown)


unflxw:
Rebased onto `wip-otel-rework`, which was rewritten when #1550, #1551, #1552, #1553 and #1554 landed on it. Before the rebase this pull request compared against the wrong merge base and appeared to change 188 files.

The two commits applied without conflicts, and the full test suite passes on the new base.

I also dropped the changeset here. The collector mode work is staying under the single changeset on `wip-otel-rework` until the whole feature is done, and this is another part of that same unreleased work.

This still needs a human review, which is why it was not included in the batch that just landed.

PR #1561, opened 2026-07-30
Skip trace context when enqueue events are off

unflxw:
This fixes behaviour that didn't match the OpenTelemetry Semantic Conventions for Messaging Spans (it would link due to messaging spans, but link to the current OpenTelemetry context at enqueue time, not to the producer span, due to the creation of the producer span/event having been disabled by a config option) which should also make it have more coherent behaviour in relation to the changes in https://github.com/appsignal/appsignal-collector/pull/418 as currently implemented (to put it in the collector PR's terms: if a trace should be kept around until it `has_links`, then it also `has_messaging`; if it doesn't `has_messaging` because of the config option that disables that, then it also won't `has_links` later on)

---

When the `enable_job_enqueue_instrumentation` option is off, the job
enqueue integrations recorded no enqueue event but still wrote trace
context onto the outgoing job. There was no producer span to point at,
so what they wrote was the context of whatever span happened to be
current, such as the surrounding web request. The job that performed
later would then link back to a span that is not a producer, which the
OpenTelemetry messaging conventions do not allow.

Each enqueue integration now checks the option itself and skips the
enqueue entirely, so it writes no trace context either. Jobs enqueued
while the option is off start their own trace.

`Transaction#job_enqueue_events_suppressed?` no longer folds the option
in, so it answers only whether an outer integration such as Active Job
is already recording this enqueue. Those nested integrations must keep
propagating trace context. The outer integration's producer span is what
the performing job links back to, and only the nested integration owns
the carrier that the job travels on.

[skip changeset]

### DIFF
--- a/lib/appsignal/hooks/active_job.rb
+++ b/lib/appsignal/hooks/active_job.rb
@@ -159,10 +159,15 @@ module Appsignal
         # transparent pass-through when there's no active transaction, and
         # `inject_context` no-ops outside collector mode.
         def enqueue(*, **)
-          # Skip recording the event when enqueue events are suppressed. That is
-          # the case when enqueue instrumentation is disabled, and it keeps this
-          # integration consistent with the standalone adapters (Sidekiq, ...),
-          # which already gate their own enqueue event on this check.
+          # When enqueue instrumentation is disabled, drop the trace context
+          # along with the event. Without an enqueue event there is no producer
+          # span, so the context we would write is that of whatever span is
+          # current, such as the surrounding web request. The job that performs
+          # later would then link back to a span that is not a producer.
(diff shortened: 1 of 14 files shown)


lipskis (approved):

unflxw:
This is merged. The commit is now on `wip-otel-rework` as [`b586eaf3`](https://github.com/appsignal/appsignal-ruby/commit/b586eaf393aa9ed8f0e34dfe2095cb92ace76537).

It was rebased onto `wip-otel-rework` rather than merged with a merge commit, so the commit has a new hash. GitHub only closes a pull request as merged when it sees the exact commits from the head branch on the base branch. That is why this one stayed open, so I am closing it by hand.

The change ships as part of #1535.

PR #1563, opened 2026-08-03
Rebuild the Excon integration

unflxw:
This does for Excon what #1523 did for Faraday: it turns the integration into
something the OpenTelemetry work can build on. The OpenTelemetry side, meaning
the client span kind, the trace context propagation and the span attributes,
follows separately on #1566.

The reason it needs rebuilding is that Excon's own instrumentor cannot measure a
request. A slow Excon request has been reported as taking almost no time.

### [Add an Excon gemfile to the test matrix](https://github.com/appsignal/appsignal-ruby/commit/298da4e93f8612ee40cf424464916102f1cf275c)

The Excon integration has never been tested against the real Excon gem.
Its spec replaces Excon with a stub class and calls the instrumentor by
hand, so nothing checks that Excon calls us the way the integration
expects. The only place the real gem is installed today is the faraday-2
gemfile, which needs it for Faraday's Excon adapter.

Add a gemfile for it and run it in CI. The gem is not pinned to a
version. Excon 1.6 needs Ruby 3.1 and Excon 1.0 needs Ruby 2.7, so the
Ruby versions already in the matrix resolve a spread of Excon versions
between them.

### [Test the Excon integration against the real gem](https://github.com/appsignal/appsignal-ruby/commit/7b149141d91850eebe6698e9038902e433ebec32)

Excon calls an instrumentor three times per request, and AppSignal
records each call as its own event. Nothing tested that against Excon
itself. The hook spec stubbed Excon out and called the instrumentor by
hand, feeding it data of its own making, so it could not tell whether
Excon calls us the way the integration expects.

Add an integration spec that drives real Excon requests, and record what
the integration produces today. Three things it produces are wrong.

The response event has no title. Excon gives an instrumentor only the
response when it reads one, and a response holds no host, so the title
the integration builds from the host has never been there. The old hook
spec passed a host in itself, which is why it never noticed.

The failure event is titled " ://". Excon gives an instrumentor only the
error when a request fails, and the integration builds a title out of a
request that is not there.

The time spent waiting for the remote service is recorded on neither
event. Excon runs its middleware stack twice, once to send the request
and again to read the response, and the middleware that reads the
response off the socket sits outside the one that calls the
instrumentor. So an Excon request is reported as taking almost no time,
however slow it really was.

A retried request records an event per attempt and one for the failure,
and a redirected request records an event per hop, so one request can
become four or more unrelated events.

Move the instrumentation tests out of the hook spec, which now covers
what the other hook specs cover: whether the dependency is there, and
whether installing works.

The Faraday spec's claim that Excon is not recorded a second time under
Faraday was passing whatever happened. The matcher it used fills in an
empty title for the keys it is not given, and a real Excon event has a
title, so the claim ruled out nothing. It now asserts on the whole list
of event names.

### [Record an Excon request as one event](https://github.com/appsignal/appsignal-ruby/commit/0a0f098527fa435a4830394fcec4364bbbf2596c)

Excon reports a request to an instrumentor in pieces: one when it sends
the request, one when it reads the response, and one when the request
fails. AppSignal recorded each piece as its own event, so one request
became two or more events, and none of them measured the request.

None of them could. Excon runs its middleware stack twice, once to send
the request and again to read the response, and the middleware that
reads the response off the socket sits outside the one that calls the
instrumentor. So the wait for the remote service falls outside every
piece an instrumentor is told about.

Instrument the connection instead. Excon::Connection#request sends the
request, waits for the response, reads it and only then returns, so one
event around that method covers the whole request and lasts as long as
the request did.

Excon retries a request, and follows a redirect, by calling that same
method again from inside the request it is retrying or following. Those
calls are suppressed with the flag the Faraday integration already
uses, so they count towards the event of the request that started them
rather than becoming events of their own.

AppSignal also no longer registers itself as Excon's instrumentor. Excon
allows only one, so an application that set up its own was having it
replaced.

### [Report the Excon integration as enabled](https://github.com/appsignal/appsignal-ruby/commit/1c65693af7e3c0a7fd098225f80ac43c35520842)

Every other HTTP client hook reports itself once it has installed, so
AppSignal knows the integration is active and not merely that the gem is
there. The Excon hook never did, so an application using Excon looked
like one that happened to have Excon installed.

### DIFF
--- a/.changesets/record-an-excon-request-as-one-event.md
+++ b/.changesets/record-an-excon-request-as-one-event.md
@@ -1,0 +1,22 @@
+---
+bump: minor
+type: change
+---
+
+Fix Excon requests being reported as taking almost no time. An Excon request was
+recorded as two events, one for sending the request and one for reading the
+response, and neither of them covered the wait for the remote service. So a slow
+Excon request looked fast in the event timeline, however long it really took.
+
+An Excon request is now recorded as a single `request.excon` event covering the
+whole request, so the event lasts as long as the request did. A request that
+Excon retried, or that was redirected, is also one event, covering every attempt
+or every hop.
+
+The `response.excon`, `retry.excon` and `error.excon` events no longer exist.
+Excon does not tell an instrumentor which request those events belonged to, so
+`response.excon` never had a title and `error.excon` was always titled `" ://"`.
+
+AppSignal also no longer registers itself as Excon's instrumentor. Excon allows
+only one instrumentor, so an application that set up its own was having it
+replaced. Your own instrumentor now keeps working.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@
 # This is a generated file by the `rake build_matrix:github:generate` task.
 # See `build_matrix.yml` for the build matrix.
 # Generate this file with `rake build_matrix:github:generate`.
-# Generated job count: 248
+# Generated job count: 255
 ---
 name: Ruby gem CI
 'on':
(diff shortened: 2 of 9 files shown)


unflxw:
Motivation for this work:

- The integration is broken. Its tests stub behaviour that the real Excon never had, and it emits broken event names when running against real Excon's data.
- The shape of events it emits isn't great. It emits two zero-length `request.excon` and `response.excon` events, rather than a single `request.excon` event with a non-zero length covering the time between the request and the response.
- That same shape of events is awkward for our present/future OpenTelemetry work -- the OpenTelemetry semantic conventions expect the request emitted and the response received to be represented as a single client span, and trace context propagation builds on that expectation. [Our collector's trace-aware-sampling early-dropping logic](https://github.com/appsignal/appsignal-collector/pull/418) will likely expect client spans with HTTP attributes to potentially propagate their context and await a future server sub-trace to arrive -- this would be true of the request span, but not of the response span.

lipskis (approved):

PR #1564, opened 2026-08-03
Fix the negated include_event matcher

unflxw:
Fix the negated include_event matcher

The include_event matcher fills in a default for every key the caller
leaves out. In the negated form that made the expectation pass almost
regardless: an event ruled out by name alone was only matched if its
title was empty too, and nearly every event has a title. So a spec
saying an event was not recorded was saying nothing, and would have gone
on passing if that event appeared.

Match on the given keys alone when the matcher is negated. The Faraday
spec is the one that was leaning on this. Its claim that Net::HTTP is
not recorded a second time under Faraday now fails if the suppression is
taken away, where before it passed either way.

Iterating a Rack response body's Enumerator turns out to record an event
that a spec said it did not. It reads the body through the same
instrumented each, so it is recorded exactly as passing a block is, and
that spec now says so. Its collector mode counterpart was written around
the old matcher's behaviour, so it says the same thing now as well.

[skip changeset]

### DIFF
--- a/spec/lib/appsignal/rack/body_wrapper_spec.rb
+++ b/spec/lib/appsignal/rack/body_wrapper_spec.rb
@@ -120,30 +120,28 @@ describe Appsignal::Rack::BodyWrapper do
         expect { |b| enum.each(&b) }.to yield_successive_args("a", "b", "c")
       end
 
+      # Iterating the returned Enumerator reads the body through the same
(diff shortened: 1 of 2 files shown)


unflxw:
Folded into #1566 as a commit. Keeping the Excon work in one pull request rather than a stack of small ones.

PR #1565, opened 2026-08-04
Test Excon suppression under Faraday again

unflxw:
Test Excon suppression under Faraday again

Faraday runs on a downstream HTTP client that AppSignal may instrument
too, and it suppresses that client so the request is recorded once. Excon
is one of the two clients this applies to, and it suppresses through a
different integration than Net::HTTP does, so it needs its own test.

That test was lost when Excon moved to its own gemfile. Excon had gained
coverage of its own by then, which is what the move was for, but nothing
was left covering the two of them together.

Put Excon back in the Faraday gemfile and test the pairing again, in both
modes. Removing Faraday's suppression now fails four examples rather than
two.

Stacked on #1564, so this targets that branch and will retarget to `wip-otel-rework` once it merges.

[skip changeset]

### DIFF
--- a/gemfiles/faraday-2.gemfile
+++ b/gemfiles/faraday-2.gemfile
@@ -1,5 +1,7 @@
 source "https://rubygems.org"
 
+gem "excon"
 gem "faraday", "~> 2.0"
+gem "faraday-excon"
 
 gemspec :path => "../"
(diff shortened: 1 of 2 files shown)


unflxw:
Folded into #1566 as a commit. Keeping the Excon work in one pull request rather than a stack of small ones.

PR #1566, opened 2026-08-04
Add OpenTelemetry support for Excon integration

unflxw:
The Excon integration is being rebuilt in #1563, against `main`. It now
instruments `Excon::Connection#request` rather than registering itself as Excon's
instrumentor, because an instrumentor is told about a request in pieces and none
of them covers the wait for the response.

This gives that new integration its OpenTelemetry behaviour: the client span
kind, the instrumentation scope, trace context propagation, and the span
attributes the semantic conventions ask for. The first three it had before and
they move onto the new instrumentation. The attributes are new, and they are the
thing the rebuild makes possible, because the request and the status of its
response now land on the same span.

It also carries two test fixes found along the way. One is a matcher that made
negated event expectations pass whatever happened. The other restores coverage of
Excon running under Faraday, which was lost when Excon moved to its own gemfile.

The first commit is temporary. It carries #1563's work so the rest can build on
it, and once #1563 lands on `main` and this branch is rebased it has nothing left
to say and drops out. Based on #1562, so it targets that branch and will retarget to `wip-otel-rework` once it merges. Merge after #1563 as well.

### [Bring in the rebuilt Excon integration](https://github.com/appsignal/appsignal-ruby/commit/c7df85a8e6c924674811589259916a93e719a717)

The Excon integration is being rebuilt in #1563, against main. It now
instruments Excon::Connection#request rather than registering itself as
Excon's instrumentor, because an instrumentor is told about a request in
pieces and none of them covers the wait for the response.

This carries that work here so the OpenTelemetry work can build on it.
The commit is temporary: once #1563 lands on main and this branch is
rebased, it has nothing left to say and drops out.

Excon loses its collector mode behaviour here, because the instrumentor
that carried it is gone. The commits that follow give it back on the new
integration, which is the point of this branch.

### [Fix the negated include_event matcher](https://github.com/appsignal/appsignal-ruby/commit/860f3856bd218a13aa088a2e8c9eeb1bd80366d9)

The include_event matcher fills in a default for every key the caller
leaves out. In the negated form that made the expectation pass almost
regardless: an event ruled out by name alone was only matched if its
title was empty too, and nearly every event has a title. So a spec
saying an event was not recorded was saying nothing, and would have gone
on passing if that event appeared.

Match on the given keys alone when the matcher is negated. The Faraday
spec is the one that was leaning on this. Its claim that Net::HTTP is
not recorded a second time under Faraday now fails if the suppression is
taken away, where before it passed either way.

Iterating a Rack response body's Enumerator turns out to record an event
that a spec said it did not. It reads the body through the same
instrumented each, so it is recorded exactly as passing a block is, and
that spec now says so. Its collector mode counterpart was written around
the old matcher's behaviour, so it says the same thing now as well.

### [Test Excon suppression under Faraday again](https://github.com/appsignal/appsignal-ruby/commit/4dc7f9106e896e7ecabaaf93eed5a5d02c8b8cb4)

Faraday runs on a downstream HTTP client that AppSignal may instrument
too, and it suppresses that client so the request is recorded once.
Excon is one of the two clients this applies to, and it suppresses
through a different integration than Net::HTTP does, so it needs its own
test.

That test was lost when Excon moved to its own gemfile. Excon had gained
coverage of its own by then, which is what the move was for, but nothing
was left covering the two of them together.

Put Excon back in the Faraday gemfile and test the pairing again, in
both modes. Removing Faraday's suppression now fails four examples
rather than two.

### [Mark Excon spans as client spans](https://github.com/appsignal/appsignal-ruby/commit/d47322660ac1f8aeb75214fdd838976f4802ea2d)

In collector mode a span says what kind of operation it covers, and an
outgoing HTTP request is a client span. It also says which
instrumentation recorded it, so an Excon request is attributed to
appsignal-ruby/excon rather than to the gem as a whole.

The Excon integration carried both of these before it was rebuilt. They
move onto the new instrumentation here, and the spec moves with them:
every example now runs in both modes, so the collector mode half can say
what the span looks like where the agent mode half says what the event
looks like.

### [Inject trace context into Excon requests](https://github.com/appsignal/appsignal-ruby/commit/ebd06e797cfc5e32624cc0b14fa23310f7876efd)

In collector mode an outgoing HTTP request carries the current trace
context, so the service being called continues the same trace instead of
starting one of its own. Excon exposes the request's headers to a
middleware, so that is where the context is written.

The Excon integration did this before it was rebuilt. What it injects
from has changed. It used to reflect the event span of one of Excon's
instrumentor notifications, and now reflects the one span that is open
around the whole request. A retried request shows the difference: every
attempt shares that span, so every attempt carries the same context.

### [Describe Excon requests on their span](https://github.com/appsignal/appsignal-ruby/commit/d843c047308259ff9c9d57c352d91f0b483fe58b)

In collector mode a span carries the attributes the OpenTelemetry
semantic conventions ask for, and for an outgoing HTTP request those are
the request method, the host and port being called, the URL, and the
status of the response.

Excon could not describe a request this way before it was rebuilt.
Everything except the status came from one of Excon's instrumentor
notifications and the status came from another, so the two never met on
the same span, and the span carrying the status carried nothing else. A
request is one span now, so all of it lands together.

A failed request is described as well, with the kind of error that ended
it. That comes from the instrumented block, which now wraps the whole
request, so it needed no work here beyond a test.

A pipelined request is the one case where Excon returns before its
response is read, so it has no status to describe. It hands back the
request data rather than a response, which is worth handling rather than
calling a method on it that is not there.

[skip changeset]


unflxw:
This is merged. Five of the six commits are now on `wip-otel-rework`:

- [`68e4a56d`](https://github.com/appsignal/appsignal-ruby/commit/68e4a56d9ba7127e02f75212b2a03a95807a16de) Fix the negated include_event matcher
- [`8b8976ab`](https://github.com/appsignal/appsignal-ruby/commit/8b8976abce0716c6bafa4714825f5b658a7f1003) Test Excon suppression under Faraday again
- [`00bf039c`](https://github.com/appsignal/appsignal-ruby/commit/00bf039c7248f926e1bf3ec1e19de0b63b9c5120) Mark Excon spans as client spans
- [`b4496214`](https://github.com/appsignal/appsignal-ruby/commit/b44962143c8227068a1b1aeae6cbda28c3b60683) Inject trace context into Excon requests
- [`cbfdc577`](https://github.com/appsignal/appsignal-ruby/commit/cbfdc57716b151383cc57227ce5fdb49fdbe0107) Describe Excon requests on their span

The sixth commit, "Bring in the rebuilt Excon integration", was left out on purpose. It brought the rebuilt Excon integration onto this branch, and that integration has since landed on `main` through #1563. So `wip-otel-rework` already has it, and applying it again would have been a no-op.

These commits are what made the rebase of `wip-otel-rework` onto `main` work. The stack had built its Excon work on the old instrumentor, which the rebuild replaced, so those changes no longer applied. The five commits here rebuild the same behavior on top of the new integration, which is exactly what the rebase needed. The Excon files on `wip-otel-rework` are now byte for byte what this branch had.

The commits were rebased rather than merged with a merge commit, so they have new hashes. GitHub only closes a pull request as merged when the base branch contains the exact commits from the head branch. That is why this one stayed open, so I am closing it by hand.

The changes ship as part of #1535.

PR #1568, opened 2026-08-07
Express per-event treatment through event formatters

unflxw:
Stacked on the OpenTelemetry work.

Two commits here are cherry-picked from pull requests against `main`, because commits in this one build on them. `Name ROM query events after ROM` is #1569, and `Title collection and layout render events` is #1570. Both touch files this pull request also changes, so the conflicts are resolved here rather than at merge time. Both go away when those land and this is rebased.

The commits before this one gave named events their own treatment by writing
them into the ActiveSupport::Notifications integration: a list of events that
are outgoing calls, a table of OpenTelemetry attributes, a lookup for the
attributes that have to be read out of a payload, and a list of events to skip.
All four are keyed by event name, and all four describe Rails components that
integration otherwise knows nothing about.

The event formatter registry is already a table keyed by event name, and it
exists so that a specific kind of event can be given specific treatment. It
could only express one kind of treatment, which is what an event's title and
body should be. These commits widen what a registration can say, and move the
four tables into it.

### [Remove an event formatter that never ran](https://github.com/appsignal/appsignal-ruby/commit/4b1ed648)

The ViewComponent formatter was registered for two event names,
`render.view_component` and `!render.view_component`. The second one
could never be reached. Events whose name starts with an exclamation
mark are internal to Rails, and this gem has skipped them since 2016,
before the registry is ever consulted. That is two years before the
registration was added.

Nothing else can reach it either. Only two places look a formatter up by
name. One skips the exclamation mark names, and the other builds the
name it looks up from a dry-monitor event, which cannot start with one.

Its only test asserted that the registration existed, so it proved
nothing beyond itself. Both are gone.

### [Let a formatter describe an event with no title](https://github.com/appsignal/appsignal-ruby/commit/0076bfa7)

An event formatter says what an event's title and body should be. That
is the only thing it can say today, so implementing `format` is
mandatory: a class without it is refused at registration.

The commits that follow give a formatter other things it can say about
an event, such as what kind of work the event is. Some events have
something to say that way but no title or body to give, so requiring a
`format` that returns nothing would be noise.

The base class implements it and returns nothing, which a formatter with
no title or body to give can now inherit. A formatter that does not
inherit from the base class still has to implement it, which is the case
the check was really written for.

### [Let a formatter declare its span kind](https://github.com/appsignal/appsignal-ruby/commit/1f90a62e)

Two places record an event that some other library reported, and both
looked up what kind of work the event was by name. The
ActiveSupport::Notifications integration kept a list of three event
names that are outgoing calls to a datastore. The dry-monitor
integration checked whether the event was called `sql`.

Those two say the same thing about the same kind of work, in two files,
because ROM reports its queries through dry-monitor while Rails reports
its own through ActiveSupport::Notifications. Both lists have to be
edited whenever a formatter is added, by someone who has to know the
lists exist.

The event formatter already exists to give a named event its own
treatment, so let it declare the kind. Both integrations now ask the
registry, and both lists are gone.

### [Let a formatter describe its own span](https://github.com/appsignal/appsignal-ruby/commit/78a73834)

The ActiveSupport::Notifications integration held a table of
OpenTelemetry attributes by event name, and a second lookup for the
attributes it has to read out of the event's payload. Both described
events that belong to Rails components the integration otherwise knows
nothing about.

Each formatter now describes its own event. The Elasticsearch one gains
the most: the index a search ran against is read from the same payload
field the formatter already reads to sanitize the search, so the two now
sit next to each other.

Performing an Active Job had attributes but no formatter, because it had
nothing to name. It gets one that says which queue the job came from and
gives the event no title, which is what it had before.

The render events only have a title inside a Rails application, because
the title is the template's path made relative to the application's
root. Whether there is an application to take a root from is now
answered by the formatter, as it formats an event, rather than by the
code that registers it. That is the formatter's own concern. Answering
it at registration time was also wrong: AppSignal can be required before
Rails is, and the answer was then wrong for as long as the process ran.

### [Let a formatter defer to another integration](https://github.com/appsignal/appsignal-ruby/commit/f02ab796)

Two events are recorded by a dedicated AppSignal integration that
describes them better than the generic paths can. The
ActiveSupport::Notifications integration held their names in a list, so
it had to know about two integrations it has nothing else to do with,
and explain both of them from a distance.

They are now claimed through the registry, the same way every other
treatment of a named event is. Both generic paths ask it, so they answer
the question the same way.

Each event is claimed by the integration that records it, as that
integration installs. The claim exists to stop the same work being
recorded twice, so it lasts exactly as long as the integration that
would otherwise record it twice. An application that turns Active Job or
Faraday instrumentation off now gets the notification that library
reports itself, where before it got no event at all.

### [Name ROM query events after ROM](https://github.com/appsignal/appsignal-ruby/commit/5e0f2e86)

A ROM query was reported as `query.postgres`, or `query.sqlite`, or
whatever else the application's database was. AppSignal groups events by
the part of the name after the dot, so this put an application's queries
in a group per database engine. The same application reported one group
in production and another in its test suite, and two applications
reaching the same database through ROM and through Sequel ended up in
unrelated groups.

The database is an attribute of the query, not its identity. ROM reports
it in the event's payload as Sequel's database type, and this used it as
the name. The name is now `query.rom` whichever database it was.

dry-monitor reports an event under an id, such as `sql`, rather than a
name. So the first value a formatter returns is what names the event
here, which is why this is the formatter's decision. An event with no
formatter was named after its id alone, which left it with no group at
all, and is now named after its id in the dry-monitor group.

The title was the event name repeated, so it is dropped. It showed the
same string twice.

### [Let a formatter declare its instrumentation scope](https://github.com/appsignal/appsignal-ruby/commit/044e630d)

The instrumentation scope names the library an event was recorded for.
Both paths that record an event some other library reported derived it
from where the event arrived, which is right until the two differ.

They differ for ROM. It reports its queries over dry-monitor, which is a
notification bus rather than something that queries a database, so every
ROM query was attributed to dry-monitor. Naming the bus says nothing
about what the query was.

A formatter can now name the library, and both paths use that when there
is one. Everything that declares nothing keeps the scope its path
derives, which is what Rails reports through
ActiveSupport::Notifications and what anything else on dry-monitor's bus
gets.

### [Title collection and layout render events](https://github.com/appsignal/appsignal-ruby/commit/22ab6ea9)

Rendering a template or a partial is titled with the template's path,
made relative to the application's root. Rendering a collection or a
layout was not titled at all, so every collection render in an
application was recorded as one event however many different partials it
rendered, and so was every layout render.

Action View reports the template's path for all four of these events, so
there was nothing to work out. These two were simply never registered
with the formatter that reads it.

Registering them means an application sees one event per template where
it saw one event in total. That is the point: it is what makes it
possible to tell which collection is the slow one.

A collection rendered from a heterogeneous set has no single template to
name, and Action View reports no path for it. The formatter already
returns no title when there is no path, so those keep the title they
have now, which is none.


lib/appsignal/integrations/active_support_notifications.rb
unflxw, line 39: This should fall back to `appsignal-ruby/active_support`, just like `dry_monitor` does.

lib/appsignal/integrations/dry_monitor.rb
```diff
@@ -4,25 +4,43 @@ module Appsignal
...
-        super
-      ensure
-        title, body, body_format = Appsignal::EventFormatter.format("#{event_id}.dry", payload)
+        begin
+          Appsignal::Transaction.current.start_event(
+            :opentelemetry_kind => Appsignal::EventFormatter.opentelemetry_kind(name),
+            :opentelemetry_scope =>
+              Appsignal::EventFormatter.opentelemetry_scope(name) ||
+                ["appsignal-ruby/dry_monitor", Appsignal::VERSION]
+          )
```
unflxw, line 27: We use `.start_event` and then `.finish_event`, both here and in `active_support_notifications`, but nothing happens in between. Let's use `.record_event` to simplify both and so the event gets correctly recorded as a zero-length event. This is a pre-existing issue, but we can fix it while we're here.

unflxw:
This is merged. Six of the eight commits are now on `wip-otel-rework`:

- [`636f1379`](https://github.com/appsignal/appsignal-ruby/commit/636f1379b219dfc262937dc4a1c911c6059a2db6) Remove an event formatter that never ran
- [`3540b6c3`](https://github.com/appsignal/appsignal-ruby/commit/3540b6c3bf950b1ce90ec0967f7b29bebd7ed9d6) Let a formatter describe an event with no title
- [`d2c119db`](https://github.com/appsignal/appsignal-ruby/commit/d2c119dbe913db6489ac1af8f5c742b861883e76) Let a formatter declare its span kind
- [`c20f2ce2`](https://github.com/appsignal/appsignal-ruby/commit/c20f2ce2f457b2a755691732e3d2442131fc1d71) Let a formatter describe its own span
- [`1ac7146e`](https://github.com/appsignal/appsignal-ruby/commit/1ac7146e7d51c34f84271954390d7d78843ee3ab) Let a formatter defer to another integration
- [`d012bd74`](https://github.com/appsignal/appsignal-ruby/commit/d012bd740bb1e45b49c519131e709549c6007f93) Let a formatter declare its instrumentation scope

The other two commits are gone on purpose. "Name ROM query events after ROM" and "Title collection and layout render events" have since landed on `main` through #1569 and #1570, so `wip-otel-rework` now gets them from `main` instead. Rebasing dropped both, because their content was already upstream.

Dropping them was not quite automatic, because each carried a little more than its counterpart on `main` does. They also updated the collector mode assertions, which `main` has no equivalent for, and the ROM one adapted to a `dry-monitor` integration that this branch had already reworked. That extra part now sits in the commits that introduce the behaviour it describes, which is where it belongs. The result is that the collector mode specs assert the ROM naming from the moment they are written, rather than asserting the old naming and being corrected later.

One conflict was worth more than a mechanical resolution. `main` registers the collection and layout render events inside its `if defined?(Rails)` guard, while "Let a formatter describe its own span" removes that guard, because the formatter now decides for itself whether there is an application to take a root from. Both are kept: all four render events are registered, and they are registered unconditionally.

The commits were rebased rather than merged with a merge commit, so they have new hashes. GitHub only closes a pull request as merged when the base branch contains the exact commits from the head branch. That is why this one stayed open, so I am closing it by hand.

The changes ship as part of #1535.

PR #1569, opened 2026-08-07
Name ROM query events after ROM

unflxw:
A ROM query was reported as `query.postgres`, or `query.sqlite`, or
whatever else the application's database was. AppSignal groups events by
the part of the name after the dot, so this put an application's queries
in a group per database engine. The same application reported one group
in production and another in its test suite, and two applications
reaching the same database through ROM and through Sequel ended up in
unrelated groups.

The database is an attribute of the query, not its identity. ROM reports
it in the event's payload as Sequel's database type, and this used it as
the name. The name is now `query.rom` whichever database it was.

dry-monitor reports an event under an id, such as `sql`, rather than a
name. So the first value a formatter returns is what names the event
here, which is why this is the formatter's decision. An event with no
formatter was named after its id alone, which left it with no group at
all, and is now named after its id in the dry-monitor group.

The title was the event name repeated, so it is dropped. It showed the
same string twice.

### DIFF
--- a/.changesets/name-rom-query-events-after-rom.md
+++ b/.changesets/name-rom-query-events-after-rom.md
@@ -1,0 +1,21 @@
+---
+bump: patch
+type: change
+---
+
+Name ROM query events after ROM, rather than after the database they ran
+against.
+
+A query made through ROM was reported as `query.postgres`, or `query.sqlite`,
+or whatever else the application's database was. The part after the dot is what
+AppSignal groups events by, so an application's queries were split into a group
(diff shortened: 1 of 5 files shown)


unflxw:
For context, this is another change towards enabling OpenTelemetry, that just happens to implement better behaviour for an obscure integration -- in this case, this allows #1568 to make agent mode and collector mode consistent between category group names and instrumentation scopes.

It's a small enough integration, with a strange enough behaviour at the moment, that I deemed it worth the continuity breakage in the data that is emitted by it.

PR #1572, opened 2026-08-11
Load the Rails app before running demo

### LINKED ISSUE #1571: `appsignal demo` does not load the Rails app, so a `config/appsignal.rb` that references Rails fails
## Describe the bug

`appsignal diagnose` loads the host Rails application before reading configuration.
`appsignal demo` does not. Both read the same `config/appsignal.rb`, so a config file
that references `Rails` works under one command and raises `NameError: uninitialized
constant Rails` under the other.

- `Appsignal::CLI::Diagnose` calls `require_rails_app_if_present`
  (`lib/appsignal/cli/diagnose.rb:635-648`), which requires the railtie and the app's
  `config/environment.rb`.
- `Appsignal::CLI::Demo` has no equivalent — `lib/appsignal/cli/demo.rb:26` calls
  `Appsignal.start` after `lib/appsignal/cli.rb` has required only `appsignal`.

`Appsignal.load_dsl_config_file` (`lib/appsignal.rb:556-585`) then rescues the error and
sets `config[:active] = false`, so the last thing printed is a generic "Unable to start
the AppSignal agent" and a recommendation to run `diagnose` — which is the one command
that does work. The actual cause appears earlier, on stderr, and is easy to miss above
the failure text.

This is not the common path, which is presumably why it has gone unnoticed: the
installer's generated config file (`resources/appsignal.rb.erb`) references no Rails
constant, and `activate_if_environment` and `env?` exist so that most files need none.
The gap is that neither helper can express *the deployment tier the application itself
resolved*. An app with two deployed tiers under one `RAILS_ENV` — a staging and a
production app both running `RAILS_ENV=production`, which is the standard Heroku
arrangement — cannot let AppSignal detect its environment, because detection reads
`Rails.env` and would label both `production`. Passing the app's own resolved value to
`Appsignal.configure` is the documented way to do that, and it is what breaks `demo`.

## To Reproduce

- Using AppSignal for Ruby gem version 4.10.0
- In my app using Rails 8.1.3.1, Ruby 4.0.6
- With this `config/appsignal.rb`:

  ```ruby
  # Anything that reaches for a Rails constant reproduces this. Here the app
  # resolves its own deployment tier, because both deployed apps run
  # RAILS_ENV=production and only the app can tell them apart.
  Appsignal.configure(Rails.application.config.x.deployment_environment) do |config|
    config.active = !Rails.env.test? && ENV["APPSIGNAL_PUSH_API_KEY"].present?
    config.name = "shipping-tracker"
  end
  ```

- Then:

  ```
  $ bundle exec appsignal demo
  appsignal ERROR: Not starting AppSignal because an error occurred while loading the AppSignal config file.
  File: "/path/to/app/config/appsignal.rb"
  NameError: uninitialized constant Rails
  Sending demonstration sample data...

  Error: Unable to start the AppSignal agent and send data to AppSignal.com.
  Please use the diagnose command (https://docs.appsignal.com/ruby/command-line/diagnose.html) to debug your configuration:

      bundle exec appsignal diagnose --environment=production
  ```

  while the same file loads correctly under `diagnose`:

  ```
  $ bundle exec appsignal diagnose --no-send-report
    environment: "development"
    name: "shipping-tracker"
  ```

The `demo` failure is identical whether or not a push API key is present, so it also
looks exactly like a credential problem to anyone provisioning a new app — which is
when `demo` is most likely to be run.

## Suggested fix

Give `Demo` the same `require_rails_app_if_present` step `Diagnose` already has. The
machinery exists one file away and is already guarded by `rails_present?`.

Two smaller improvements stand on their own if the above is not wanted:

- Have the config-file rescue name the file's own error in the *final* message rather
  than only in a preceding stderr line. As it stands the last thing on screen sends the
  reader to `diagnose` to discover a cause the tool has already established.
- Document that the Ruby config file may be loaded without the framework booted, so it
  should not assume the application is available. Nothing in the configuration docs
  suggests the file is read in two different environments.

## Related

#1548 notes from the other direction that `diagnose` reports a wrong or missing push API
key while `demo` does not. Same asymmetry between the two commands, different cause —
this one is that `demo` never boots the framework at all, so the fix does not overlap.

Guflly:
`appsignal demo` now boots a Rails application before transmitting, matching `appsignal diagnose`. This lets `config/appsignal.rb` reference Rails state and passes the selected environment through while the app boots.

Both commands use the same loader, which also clears the temporary config environment if loading fails.

Fixes #1571

### DIFF
--- a/.changesets/fix-demo-rails-app-loading.md
+++ b/.changesets/fix-demo-rails-app-loading.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+type: fix
+---
+
+Load the host Rails application before the `appsignal demo` command reads its
+configuration. This allows `config/appsignal.rb` to reference Rails application
+state, matching the behavior of `appsignal diagnose`.
--- a/lib/appsignal/cli/demo.rb
+++ b/lib/appsignal/cli/demo.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "appsignal/cli/helpers"
 require "appsignal/demo"
 
 module Appsignal
@@ -40,6 +41,8 @@ module Appsignal
     # @see https://docs.appsignal.com/support/debugging.html
     #   Debugging AppSignal guide
     class Demo
+      extend CLI::Helpers
+
       class << self
         # @param options [Hash]
         # @option options :environment [String] environment to load
@@ -47,6 +50,7 @@ module Appsignal
         # @return [void]
         def run(options = {})
           ENV["APPSIGNAL_APP_ENV"] = options[:environment] if options[:environment]
+          require_rails_app_if_present(options[:environment])
 
           puts "Sending demonstration sample data..."
           if Appsignal::Demo.transmit
@@ -63,6 +67,19 @@ module Appsignal
             puts
             exit 1
           end
+        end
+
+        private
+
+        def require_rails_app_if_present(environment)
+          return unless rails_present?
+
+          load_rails_app(environment)
+        rescue LoadError, StandardError => error
+          puts
+          puts "ERROR: Error encountered while loading the Rails app"
+          puts "#{error.class}: #{error.message}"
+          puts error.backtrace
         end
       end
     end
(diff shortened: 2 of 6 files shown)


unflxw:
Thanks for your contribution @Guflly! Looking into it 👀

unflxw:
Continuing the work in #1584.

PR #1574, opened 2026-08-18
Honor CA file environment variables during agent download

Cosmo:
## Summary

Use configured CA certificate paths when downloading the AppSignal agent during extension installation.

The installer currently always uses the bundled `resources/cacert.pem`. This causes installation to fail behind TLS-intercepting proxies such as Zscaler, even when `APPSIGNAL_CA_FILE_PATH` points to the trusted system CA bundle.

The installer now selects the CA certificate path in this order:

1. `APPSIGNAL_CA_FILE_PATH`
2. `SSL_CERT_FILE`
3. The existing bundled `resources/cacert.pem`

The existing default behavior remains unchanged when neither environment variable is configured.

## Motivation

For the past years in our company we were able to disable Zscaler while installing gems or building containers and never bothered to touch the AppSignal gem.
With recent policy changes, Zscaler is now running non-stop and without the change we can't get AppSignal running.

## Testing

- Added specs covering the configured paths, precedence, blank values, and bundled fallback.
- RuboCop passes for the changed Ruby files.
- Verified `rake extension:install` successfully downloads and installs the agent behind Zscaler when `APPSIGNAL_CA_FILE_PATH` is configured.
- Verified installation from the fork in an affected Rails devcontainer.

### DIFF
--- a/.changesets/honor-agent-download-ca-file-path.md
+++ b/.changesets/honor-agent-download-ca-file-path.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+Honor `APPSIGNAL_CA_FILE_PATH` and `SSL_CERT_FILE` when downloading the agent during extension installation.
--- a/ext/base.rb
+++ b/ext/base.rb
@@ -22,6 +22,11 @@ def ext_path(path)
   File.join(EXT_PATH, path)
 end
 
+def ca_cert_path
+  ENV.values_at("APPSIGNAL_CA_FILE_PATH", "SSL_CERT_FILE")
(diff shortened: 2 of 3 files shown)


unflxw:
Good catch! We _should_ be honoring `APPSIGNAL_CA_FILE_PATH` for the agent download, thank you for your contribution!

Not sure about `SSL_CERT_FILE`, though -- seems to be an OpenSSL internal-ish we don't necessarily want to assume is set for us, and we don't honor it in other integrations.

unflxw:
Closed in favor of #1575.

unflxw:
Released AppSignal for Ruby version 4.10.1, with a fix to this issue. Thanks again for your contribution!

PR #1580, opened 2026-08-20
Name the real SQL engine in db.system.name

unflxw:
### [Stop the sentinel overwriting db.system.name](https://github.com/appsignal/appsignal-ruby/commit/d0b7f38f)

write_event_body_attributes wrote the other_sql sentinel on every SQL
event unconditionally. finish_event and record_event both call it
after an integration's own attributes are already on the span, so the
sentinel always overwrote whatever engine name the integration had
set. Nothing in the gem sets db.system.name yet, so the bug had no
visible effect, but it would silently undo every integration change
that starts naming a real engine.

The event frame now tracks whether db.system.name was set during the
event, through set_attributes, and the sentinel is only written when
that never happened. record_event has no event frame to track this
on, so it always falls back to the sentinel, which is correct until a
later change gives it a way to pass attributes of its own.

### [Apply formatter attributes to dry-monitor events](https://github.com/appsignal/appsignal-ruby/commit/44316b73)

instrument called Appsignal::EventFormatter.format but never
Appsignal::EventFormatter.opentelemetry_attributes, unlike
ActiveSupportNotificationsIntegration#finish_event, which calls both.
A formatter that only arrives over dry-monitor, such as ROM's SQL
formatter, could define opentelemetry_attributes and it would never
run.

The call is added before finish_event, so the attributes land on the
event's own span rather than on the transaction.

### [Let record_event take opentelemetry_attributes](https://github.com/appsignal/appsignal-ruby/commit/72cab63a)

record_event already accepted opentelemetry_kind and
opentelemetry_scope, but had no way to attach attributes. instrument
callers can add attributes in the window between start_event and
finish_event, but record_event reports a complete event in one call,
so a caller like DataMapper's integration had no way to describe what
it recorded.

The new keyword threads through Transaction#record_event and
OpenTelemetryBackend#record_event, which formats and writes the
attributes to the span before it finishes. ExtensionBackend#record_event
ignores it, the same as the other OpenTelemetry keywords, since agent
mode has no span to put it on. A caller that names its own
db.system.name this way takes priority over the SQL sentinel, the same
as during a normal instrumented event.

record_event is `@!visibility private`, so this needs no signature
changes; sig/ stays clean after regenerating.

### [Name the engine for ActiveRecord, Sequel, ROM](https://github.com/appsignal/appsignal-ruby/commit/41193382)

Every SQL integration recorded a client span with no database
attributes beyond the query text, while Redis, MongoDB and
Elasticsearch already report a full set. The semantic conventions ask
for db.system.name on every database span, but until now every SQL
span got the same "we don't know the dialect" sentinel.

A new Appsignal::OpenTelemetry::SqlDbSystem module maps each library's
own name for its engine onto the semantic conventions value, since
none of the libraries spell it the same way and none of them match the
conventions either. It keeps one map per library rather than one
shared map, because the three libraries' vocabularies do not collide
today only by accident: ActiveRecord's ADAPTER_NAME strings, Sequel's
database_type symbols, and DataMapper's DataObjects class names could
easily overlap in the future, and a flat map would then return a
plausible-looking wrong answer instead of nil. ROM is the one
deliberate exception. Its dry-monitor payload already carries
Sequel's own database_type symbol, so it reuses Sequel's lookup
directly rather than a coincidentally similar one.

ActiveRecord's formatter reads payload[:connection].adapter_name,
covering the Postgres, Mysql2, SQLite and Trilogy adapters Rails
bundles, plus SQL Server and Oracle through the separate
activerecord-sqlserver-adapter and activerecord-oracle_enhanced-adapter
gems, which declare ADAPTER_NAME the same way. Sequel and DataMapper
already named SQL Server, so leaving ActiveRecord's SQL Server users
on the sentinel would have been an oversight rather than a choice.
Sequel's own hook reads database_type directly off the Sequel::Database
the query ran on, since the sql.sequel formatter (used for
sequel-rails compatibility) never sees that object, only the query
text.

This follows the older semantic conventions value set: SQL Server maps
to "mssql" and Oracle maps to "oracle", not the current registry's
"microsoft.sql_server" and "oracle.db", because those are the values
the AppSignal collector's sanitizer recognizes today. A name the map
does not recognise returns nil, so the caller's own other_sql fallback
still applies, the same as for a library this gem has no mapping for
at all.

The Sequel hook also names db.namespace from opts[:database], Sequel's
own option key for the database it connected to. It is a single
documented accessor, so it clears the bar the equivalent ActiveRecord
attribute does not: reaching a connection's database name there goes
through the connection pool's configuration, and the accessor has
moved across the Rails versions this gem supports.

### [Name the real engine for DataMapper](https://github.com/appsignal/appsignal-ruby/commit/d29625e1)

ActiveRecord, Sequel and ROM already name the real db.system.name
instead of the "we don't know the dialect" sentinel. DataMapper's
integration already checks its connection class against SQL_CLASSES
to decide whether a query is SQL at all, so the same class also names
the engine through the shared SqlDbSystem mapping.

This needed record_event's new opentelemetry_attributes keyword,
because DataMapper reports a complete event through record_event in
one call, with no window between start and finish to add an attribute
the way an instrumented event can.

attributes is built as a hash from the start and left empty when there
is nothing to add, rather than threaded through as nil on the branches
that have nothing to report. add_opentelemetry_attributes already
treats an empty hash as a no-op, so there is no need for a value that
exists only to satisfy the variable's later reference.

[skip changeset]

### DIFF
--- a/lib/appsignal/event_formatter/active_record/sql_formatter.rb
+++ b/lib/appsignal/event_formatter/active_record/sql_formatter.rb
@@ -10,6 +10,20 @@ module Appsignal
           :client
         end
 
+        # The payload carries the connection the query ran on (Rails 6.0+),
+        # whose `adapter_name` is each adapter's own name for itself, such as
+        # `"PostgreSQL"` or `"Mysql2"`. A name the mapping does not recognise
+        # is left to the SQL sentinel, same as an adapter this gem has never
+        # heard of.
+        def opentelemetry_attributes(payload)
+          name = Appsignal::OpenTelemetry::SqlDbSystem.name_for_active_record(
+            payload[:connection]&.adapter_name
+          )
+          return unless name
+
+          { "db.system.name" => name }
+        end
+
         def format(payload)
           [payload[:name], payload[:sql], SQL_BODY_FORMAT]
         end
--- a/lib/appsignal/event_formatter/rom/sql_formatter.rb
+++ b/lib/appsignal/event_formatter/rom/sql_formatter.rb
@@ -29,6 +29,18 @@ module Appsignal
         def format(payload)
           ["query.rom", payload[:query], SQL_BODY_FORMAT]
         end
+
+        # The payload's `name` is Sequel's `database_type` symbol for the
+        # database ROM is talking to, so this uses Sequel's own lookup
+        # rather than a coincidentally similar one. A symbol the mapping does
+        # not recognise is left to the SQL sentinel, same as an engine this
+        # gem has never heard of.
+        def opentelemetry_attributes(payload)
+          name = Appsignal::OpenTelemetry::SqlDbSystem.name_for_sequel(payload[:name])
+          return unless name
+
+          { "db.system.name" => name }
+        end
       end
     end
   end
--- a/lib/appsignal/hooks/sequel.rb
+++ b/lib/appsignal/hooks/sequel.rb
@@ -14,6 +14,9 @@ module Appsignal
           :opentelemetry_kind => :client,
           :opentelemetry_scope => ["appsignal-ruby/sequel", Appsignal::VERSION]
         ) do
+          Appsignal::Transaction.current.add_opentelemetry_attributes(
+            Appsignal::Hooks::SequelHook.sequel_db_attributes(self)
+          )
           super
         end
       end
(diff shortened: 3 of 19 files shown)


unflxw:
Applied to `wip-otel-rework`, rebased onto the branch tip since #1579 had landed in the meantime. The five commits are [8820a8e9](https://github.com/appsignal/appsignal-ruby/commit/8820a8e9a56aada3544701e08d13be6ceacb6b0c) through [99445be9](https://github.com/appsignal/appsignal-ruby/commit/99445be97501b30f69a642391e2ada56138379ed), and they are listed in the description of #1535.

Closing rather than merging. The rebase gave the commits new hashes, so GitHub cannot recognise them as this branch's.

PR #1581, opened 2026-08-21
Use `app` as the default service name

unflxw:
The collector turns the service name into names that customers see. A
trace's namespace becomes `{service name}/{namespace}`, so a web
request from an app that does not set `service_name` was filed under
`unknown/web`. A log line's group falls back to the service name on
its own, so those lines were grouped under `unknown`.

Leaving `service_name` unset is the normal case for an app that is a
single service. Nothing about it is actually unknown, so the fallback
now describes what it is. The `revision` and `hostname` fallbacks
still report `unknown`, because those are facts about the running
process that AppSignal really does not have.

### DIFF
--- a/.changesets/default-service-name-to-app.md
+++ b/.changesets/default-service-name-to-app.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: change
+---
+
+Report `app` instead of `unknown` as the OpenTelemetry service name when the `service_name` configuration option is not set and collector mode is in use.
--- a/lib/appsignal/opentelemetry.rb
+++ b/lib/appsignal/opentelemetry.rb
@@ -212,10 +212,12 @@ module Appsignal
 
       # Build the OpenTelemetry Resource that carries AppSignal config to the
       # collector. Attributes whose underlying option is nil or an empty array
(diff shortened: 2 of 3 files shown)


unflxw:
Rationale for this (and for consistency, its twin Python PR at https://github.com/appsignal/appsignal-python/pull/270) is to have the same name for customers who upgrade from agent mode to collector mode. @matsimitsu will be changing the processor's translation layer to use "app" as well, so we have a shared, reasonable enough default.

Docs will still advise setting a service name as part of collector mode setup, so the fallback may not be very relevant, but customers who upgrade somewhat manually may skip the docs.

unflxw:
Now that collector mode has been released in `5.0.0.rc1`, this should have a changeset so that `.rc2` can explain the change. The one from the Python twin PR would do nicely.

unflxw:
Added the changeset, copied from the Python twin PR. Rebased onto `main` so it lands on top of the `5.0.0.rc.1` release, and dropped the paragraph about there being no changeset from the commit message and the description.

matsimitsu:
(This has also been implemented in the legacy > otel mapper(s))

PR #1587, opened 2026-08-25
Send app_path and ignore_logs to the collector

### LINKED ISSUE #1585: Send app_path and ignore_logs as resource attributes in collector mode
The collector understands two configuration values that the gem never sends it. Both already reach the agent through the `_APPSIGNAL_*` environment variables, so the gem has the values. `Appsignal::OpenTelemetry.build_resource` just does not put them in the resource.

Sending them is all that is needed. Nothing has to change in the collector or the processor.

### `app_path`

The gem sets `_APPSIGNAL_APP_PATH` to `root_path` for the agent. The collector reads `appsignal.config.app_path` and, unlike the other AppSignal config attributes, keeps it on the outgoing span batch instead of stripping it. See `KEEP_APPSIGNAL_RESOURCE_CONFIG` in `collector/src/tracing/traces_buffer.rs`. The processor reads it back off the span's resource attributes and uses it to build the backtrace parser.

The app path is what gets stripped from each backtrace line. Without it, the lines keep their absolute paths, so a collector-mode app shows `/app/app/models/product.rb` where an agent-mode one shows `app/models/product.rb`. It gets worse from there, because the processor decides whether a frame belongs to the application by checking whether its path is relative. Every frame now starts with a `/`, so none of the application's own frames are recognised as app lines.

### `ignore_logs`

In agent mode the filtering happens in the extension, in `create_log_line`. In collector mode logs go through `Appsignal::Logger::OpenTelemetryBackend` instead, so that code never runs.

The collector implements the same filtering in `collector/src/logs/logs_buffer.rs`, down to the same small subset of `^`, `$` and `.*` matching. It only applies it when the resource carries `appsignal.config.ignore_logs`. Since the gem does not send that attribute, `ignore_logs` currently does nothing at all in collector mode.

The Python package has the same gap for `app_path`, tracked in https://github.com/appsignal/appsignal-python/issues/273.

It has no `ignore_logs` option at all yet, because until collector mode it sent no logs to filter. Adding one is tracked in https://github.com/appsignal/appsignal-python/issues/275.

unflxw:
Fixes https://github.com/appsignal/appsignal-ruby/issues/1585.

### [Send the app path to the collector](https://github.com/appsignal/appsignal-ruby/commit/1710aafe3062a5e43b5460cb4a0747d1b089d528)

The processor strips the app path from each backtrace line, and decides
whether a frame belongs to the application by checking that its path is
relative. Without the app path, every frame keeps its absolute path, so
no line is recognized as the application's own.

### [Send ignore_logs to the collector](https://github.com/appsignal/appsignal-ruby/commit/a9e0f202ba8ae9671e3a2503e28246e8e7227889)

In collector mode log lines go through the OpenTelemetry logger backend
instead of the extension, so the extension's filtering never runs. The
collector applies the same filtering itself, but only when the resource
carries the option.

### DIFF
--- a/.changesets/send-ignore-logs-to-the-collector.md
+++ b/.changesets/send-ignore-logs-to-the-collector.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+The `ignore_logs` option now filters out the log lines that match its patterns in collector mode.
--- a/.changesets/send-the-app-path-to-the-collector.md
+++ b/.changesets/send-the-app-path-to-the-collector.md
@@ -1,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognized as your application's code.
--- a/lib/appsignal/opentelemetry.rb
+++ b/lib/appsignal/opentelemetry.rb
@@ -226,6 +226,7 @@ module Appsignal
           "appsignal.config.push_api_key" => config[:push_api_key],
           "appsignal.config.revision" => revision,
           "appsignal.config.app_path" => config.root_path&.to_s,
+          "appsignal.config.platform" => config[:platform],
           "appsignal.config.language_integration" => "ruby",
           "service.name" => service_name,
           "host.name" => host_name,
--- a/lib/appsignal/opentelemetry.rb
+++ b/lib/appsignal/opentelemetry.rb
@@ -237,6 +237,7 @@ module Appsignal
           "appsignal.config.filter_request_session_data" => config[:filter_session_data],
           "appsignal.config.ignore_actions" => config[:ignore_actions],
           "appsignal.config.ignore_errors" => config[:ignore_errors],
+          "appsignal.config.ignore_logs" => config[:ignore_logs],
           "appsignal.config.ignore_namespaces" => config[:ignore_namespaces],
           "appsignal.config.response_headers" => config[:response_headers],
           "appsignal.config.request_headers" => config[:request_headers],
--- a/lib/appsignal/opentelemetry.rb
+++ b/lib/appsignal/opentelemetry.rb
@@ -225,6 +225,7 @@ module Appsignal
           "appsignal.config.environment" => config.env,
           "appsignal.config.push_api_key" => config[:push_api_key],
           "appsignal.config.revision" => revision,
+          "appsignal.config.app_path" => config.root_path&.to_s,
           "appsignal.config.language_integration" => "ruby",
           "service.name" => service_name,
           "host.name" => host_name,
--- a/spec/lib/appsignal/opentelemetry_spec.rb
+++ b/spec/lib/appsignal/opentelemetry_spec.rb
@@ -340,6 +340,7 @@ if DependencyHelper.opentelemetry_present?
       it "maps AppSignal config attributes onto the resource" do
         resource = described_class.build_resource(
           build_config(
+            :root_path => "/path/to/app",
             :options => {
               :name => "my-app",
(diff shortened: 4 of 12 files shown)


.changesets/send-ignore-logs-to-the-collector.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+In collector mode, the `ignore_logs` option now filters out the log lines that match its patterns. Before, it had no effect in collector mode.
```
unflxw:
```suggestion
The `ignore_logs` option now filters out the log lines that match its patterns in collector mode.
```

.changesets/send-the-app-path-to-the-collector.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognised as your application's code. Before, they were shown as absolute paths, and no line was marked as coming from your application.
```
unflxw:
```suggestion
In collector mode, backtrace lines from your own application are now shown as paths relative to your application's root, and are recognised as your application's code.
```

unflxw (commented):
No need to say "before, the opposite was true" -- it's implied from "this is now true".

.changesets/send-ignore-logs-to-the-collector.md
unflxw: Applied.

.changesets/send-the-app-path-to-the-collector.md
unflxw: Applied, with one change: I used `recognized` rather than `recognised`. Both changelogs use the American spelling everywhere else, and your suggestion on the Python pull request uses it too.

lib/appsignal/opentelemetry.rb
```diff
@@ -225,6 +225,7 @@ def build_resource(config)
           "appsignal.config.environment" => config.env,
           "appsignal.config.push_api_key" => config[:push_api_key],
           "appsignal.config.revision" => revision,
+          "appsignal.config.app_path" => config.root_path&.to_s,
```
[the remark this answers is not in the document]
unflxw: The app path is already sent to AppSignal in agent mode. The gem passes it to the agent as `_APPSIGNAL_APP_PATH`, and the agent puts it in the payload it sends. So this makes collector mode report what agent mode already reports, rather than disclosing something new.

It is also load bearing: the processor uses the app path to strip it from each backtrace line and to decide which frames belong to the application. Without it those frames are not recognised.

spec/lib/appsignal/opentelemetry_spec.rb
```diff
@@ -389,6 +393,22 @@ def with_leaked_ambient_context
...
+              :options => {
+                :name => "my-app",
+                :push_api_key => "abc"
+              }
+            )
+          )
+
+          expect(resource_attrs(resource)).not_to have_key("appsignal.config.app_path")
+        end
+      end
```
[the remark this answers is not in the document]
unflxw: Dismissed. A block parameter in Ruby is a fresh local on every iteration, so each `it` block closes over its own value. There is no shared variable to capture:

```
$ ruby -e 'procs = []; [nil, ""].each { |x| procs << -> { x.inspect } }; puts procs.map(&:call).inspect'
["nil", "\"\""]
```

The two examples run as `omits the app path when the root path is nil` and `omits the app path when the root path is ""`, and this is the same pattern the revision specs in `config_spec.rb` use to generate one example per environment variable.

lipskis (approved):

PR #1590, opened 2026-08-27
Report Delayed Job jobs that fail to load

unflxw:
### [Report Delayed Job jobs that fail to load](https://github.com/appsignal/appsignal-ruby/commit/6f2a837131dc5af7eafac7fb6cfdf69ac8f16f60)

The Delayed Job integration reads a job's payload after performing it,
to name the transaction and record the job's parameters. Delayed Job
raises when a job's handler no longer deserializes, and raises again on
every further read, so that read raises inside the `ensure` block. The
transaction is therefore never completed and the failure never reaches
AppSignal, even though the error is already set on it. The payload read
is now guarded, and a job whose payload cannot be read is named after
the class recorded in its raw handler, or after Delayed Job itself when
the handler does not record one either.

### [Report Delayed Job jobs the worker cannot load](https://github.com/appsignal/appsignal-ruby/commit/06d99afb41a60e37437e4749c45ae9dfefc49966)

The plugin instruments a job from Delayed Job's `invoke_job` callback.
Delayed Job reads a job's payload before it invokes it, because it asks
the job for its own maximum run time, so a job whose payload will not
deserialize raises inside the worker and never reaches `invoke_job`.
The worker fails that job on its own and nothing is reported. The
`perform` callback runs for every job the worker takes, including that
one, so reading the payload there is what gets the failure reported.

The transaction stays on `invoke_job`. `Delayed::Worker#run` rescues
every exception a job raises, so a transaction opened around it would
never see the job's error, and error reporting would stop for every
failing job with nothing to show that it had.

### DIFF
--- a/.changesets/report-delayed-job-jobs-that-fail-to-load.md
+++ b/.changesets/report-delayed-job-jobs-that-fail-to-load.md
@@ -1,0 +1,13 @@
+---
+bump: patch
+type: fix
+---
+
+Report Delayed Job jobs that fail to load. Before this change, a job whose
+payload could not be deserialized was not reported to AppSignal at all.
+
+This can happen when a deploy removes or renames a job class while jobs of that
+class are still queued. Those jobs are now reported as failed, with the
+`Delayed::DeserializationError` they raised, and are named after the class
+recorded in the job's handler. When even that cannot be read, they are named
+`DelayedJobInternal`.
--- a/lib/appsignal/integrations/delayed_job_plugin.rb
+++ b/lib/appsignal/integrations/delayed_job_plugin.rb
@@ -63,16 +63,26 @@ module Appsignal
           transaction.set_error(error)
           raise
         ensure
-          payload = job.payload_object
-          if payload.respond_to? :job_data
-            # ActiveJob
-            job_data = payload.job_data
-            transaction.set_action_if_nil("#{job_data["job_class"]}#perform")
-            transaction.add_params_if_nil(job_data.fetch("arguments", {}))
-          else
-            # Delayed Job
-            transaction.set_action_if_nil(action_name_from_payload(payload, job.name))
-            transaction.add_params_if_nil(extract_value(payload, :args, {}))
+          # Delayed Job raises when a job's handler will not deserialize, and it
+          # raises again on every further attempt to read it. Reading the payload
+          # here without a guard therefore replaces the error the job already
+          # failed with, and skips the rest of this block, so the transaction is
+          # never completed and the failure is never reported.
+          begin
+            payload = job.payload_object
+            if payload.respond_to? :job_data
+              # ActiveJob
+              job_data = payload.job_data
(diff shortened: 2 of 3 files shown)


.changesets/report-delayed-job-jobs-that-fail-to-load.md
```diff
@@ -0,0 +1,13 @@
+---
+bump: patch
+type: fix
+---
+
+Report Delayed Job jobs that fail to load. A job whose stored handler can no
+longer be turned back into a Ruby object was not reported to AppSignal at all,
+so the failure was invisible.
```
unflxw:
```suggestion
Report Delayed Job jobs that fail to load. Before this change, a job whose payload could not be deserialised was not reported to AppSignal at all.
```

.changesets/report-delayed-job-jobs-that-fail-to-load.md
```diff
@@ -0,0 +1,13 @@
...
+
+This happens when a deploy removes or renames a job class while jobs of that
```
unflxw:
```suggestion
This can happen when a deploy removes or renames a job class while jobs of that
```

lib/appsignal/integrations/delayed_job_plugin.rb
```diff
@@ -63,16 +63,26 @@ def self.invoke_with_instrumentation(job, block)
...
+              transaction.add_params_if_nil(job_data.fetch("arguments", {}))
+            else
+              # Delayed Job
+              transaction.set_action_if_nil(action_name_from_payload(payload, job.name))
+              transaction.add_params_if_nil(extract_value(payload, :args, {}))
+            end
+          rescue => error
+            warn_unreadable_payload_once(error)
+            transaction.set_action_if_nil(action_name_without_payload(job))
           end
```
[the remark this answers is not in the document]
unflxw: Keeping this rescue broad is deliberate, because the block runs inside `ensure`. Any exception that escapes it skips `complete_current!`, and that is the transaction leak this commit exists to fix.

Narrowing it to `Delayed::DeserializationError` would let real cases through. Delayed Job's `payload_object` only converts `TypeError`, `LoadError`, `NameError`, `ArgumentError`, `SyntaxError` and `Psych::SyntaxError` into a `DeserializationError`. A handler naming a class Psych will not permit raises `Psych::DisallowedClass`, which is a `RuntimeError` and is not in that list, so it comes out unwrapped.

The naming path can also raise on its own. `job.name` rescues `DeserializationError` and then reads the class out of the raw handler with `ParseObjectFromYaml.match(handler)[1]`, which raises `NoMethodError` when the handler does not match that expression.

On the warning reading inaccurately: it interpolates the error class and message, so the real cause is always in the log line even when it is not a deserialization failure.

tombruijn (commented):
See my comments on the v5 branch PR: https://github.com/appsignal/appsignal-ruby/pull/1592#pullrequestreview-5068297617

Issue #1578, opened 2026-08-19
perform_all_later records an adapter enqueue event per job on top of enqueue_all.active_job


jcbpl:
## Describe the bug

`ActiveJobEnqueueInstrumentation#enqueue` records one `enqueue.active_job` event and wraps the call in `suppress_job_enqueue_events`, so the adapter doesn't record its own enqueue event on top of it. The comment there says the intent is to record the enqueue once rather than as nested Active Job + adapter events.

`ActiveJob.perform_all_later` never enters `#enqueue`. Active Job calls the adapter's `enqueue_all` instead, which for Sidekiq is `push_bulk`, and Sidekiq's client middleware runs once per job with no suppression in effect. A bulk enqueue of 3 jobs records 4 events: one `enqueue_all.active_job` and three `enqueue.sidekiq`.

I expected the bulk path to record either one event per job, matching the single-job path, or one event for the batch, but not both.

A job that slices a large collection and calls `perform_all_later` per batch therefore records an event per job enqueued, where 4.8.4 recorded one per batch.

One related question, since it affects how much the duplication matters: is there a cap on events per transaction, and where does it apply? The extension keeps everything it's given—100,000 events came back from `Transaction#to_h` intact, at roughly 26MB of JSON—while samples I look at in the UI appear trimmed to the slowest events. If the agent does the trimming, a large bulk enqueue costs little beyond the events themselves. If it happens on ingest, the full payload crosses the wire first.

## To Reproduce

- Using AppSignal for Ruby gem version 4.9.1
- With activejob 8.1.3.1, sidekiq 8.1.0 and ruby 4.0.5
- In this standalone script:

```ruby
require "active_job"
require "sidekiq"
require "appsignal"

Appsignal.configure(:development) do |config|
  config.active = true
  config.push_api_key = "..."
  config.name = "repro"
end
Appsignal.start

ActiveJob::Base.queue_adapter = :sidekiq

class ExampleJob < ActiveJob::Base
  def perform(number) = nil
end

def event_counts
  transaction = Appsignal::Transaction.create(Appsignal::Transaction::BACKGROUND_JOB)
  transaction.set_action("Repro#perform")
  yield
  counts = transaction.to_h["events"].group_by { |event| event["name"] }.transform_values(&:size)
  Appsignal::Transaction.complete_current!
  counts
end

p event_counts { 3.times { |number| ExampleJob.perform_later(number) } }
[...]

PR #1591, opened 2026-08-27
Record Active Job bulk enqueues as one event

unflxw:
Fixes https://github.com/appsignal/appsignal-ruby/issues/1578.

Active Job's `perform_all_later` does not enqueue through
`ActiveJob::Base#enqueue`, so the wrapper that records an enqueue once
never runs for a batch. Nothing suppresses the queue adapter's own
enqueue instrumentation, and a batch of three jobs is therefore recorded
as Rails' native batch notification plus one adapter event per job. The
batch is now recorded as a single event, with the adapter's own events
suppressed while it is enqueued. The method wrapped to do that is
private, because it is the only point that covers both the adapters that
enqueue a batch themselves and the ones Active Job falls back to
enqueuing job by job.

The wrapper reads the batch out of that method's second argument, and
the method being private means a later version of Active Job can change
what it takes. Wrapping a method that takes something else raises an
`ArgumentError` inside every `ActiveJob.perform_all_later` call an
application makes, which breaks enqueuing and not just its
instrumentation. It is therefore wrapped only while its parameters are
the ones the wrapper was written against. A batch that cannot be
recorded that way goes unrecorded, because Rails' own notification is
the worse event this hook set out to replace.

### DIFF
--- a/.changesets/record-active-job-bulk-enqueues-as-one-event.md
+++ b/.changesets/record-active-job-bulk-enqueues-as-one-event.md
@@ -1,0 +1,14 @@
+---
+bump: patch
+type: fix
+---
+
+Record a single event for a bulk enqueue of Active Job jobs. Before this change,
+enqueuing jobs with `ActiveJob.perform_all_later` would record an event for the
+batch and, if the adapter was instrumented with AppSignal, another event for each
+job in it. A job that slices a large collection and enqueues it in batches was
+therefore reported with an event per job enqueued, where before version 4.9.0 it
+had one event per batch.
+
+A bulk enqueue is now recorded as a single `enqueue_all.active_job` event, named
+after the job class when every job in the batch shares one.
--- a/lib/appsignal/hooks/active_job.rb
+++ b/lib/appsignal/hooks/active_job.rb
@@ -21,6 +21,59 @@ module Appsignal
         defined?(::ActiveJob)
       end
 
+      # The parameters of the Active Job method this hook wraps to record a
+      # bulk enqueue. The wrapper reads the array of jobs out of the second
+      # one, so it is only safe to wrap a method that still takes these.
+      EXPECTED_INSTRUMENT_ENQUEUE_ALL_PARAMETERS = [
+        [:req, :queue_adapter],
+        [:req, :jobs]
+      ].freeze
+
+      # Whether Active Job records a bulk enqueue through a method at all. It
+      # only does so from version 7.1 on. Checking for the method, rather than
+      # for the version, keeps us from wrapping one on a version that has no
+      # bulk enqueue path to instrument.
+      def self.instrument_enqueue_all_defined?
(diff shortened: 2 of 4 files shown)


lib/appsignal/hooks/active_job.rb
```diff
@@ -33,6 +33,15 @@ def install
...
             .prepend ::Appsignal::Hooks::ActiveJobHook::ActiveJobEnqueueInstrumentation
 
+          # Active Job records a bulk enqueue through this method, which only
+          # exists from version 7.1 on. Checking for the method, rather than for
+          # the version, keeps us from defining one on a version that has no
+          # bulk enqueue path to instrument.
+          if ::ActiveJob.singleton_class.private_method_defined?(:instrument_enqueue_all)
+            ::ActiveJob.singleton_class
+              .prepend ::Appsignal::Hooks::ActiveJobHook::ActiveJobBulkEnqueueInstrumentation
+          end
```
unflxw, line 51: If the version is 7.1+ and this does _not_ instrument it, we may want to log a warning about it (something like "Could not instrument `ActiveJob.perform_all_later`, bulk enqueues will not be instrumented correctly") and a test should fail.

EDIT: a test does fail; skipping logging a warning for now

.changesets/record-active-job-bulk-enqueues-as-one-event.md
```diff
@@ -0,0 +1,16 @@
...
+`ActiveJob.perform_all_later` recorded an event for the batch and then another
+event for each job in it. A job that slices a large collection and enqueues it
+in batches was therefore reported with an event per job enqueued, where before
+version 4.9.0 it had one event per batch.
+
+A bulk enqueue is now recorded as a single `enqueue_all.active_job` event, named
+after the job class when every job in the batch shares one.
+
+This affected every queue adapter, not only Sidekiq, because a bulk enqueue does
+not go through the code path that records a single enqueue.
```
unflxw: Nothing else talks about Sidekiq or creates the impression that it would affect only some adapters.

```suggestion
```

.changesets/record-active-job-bulk-enqueues-as-one-event.md
```diff
@@ -0,0 +1,16 @@
+---
+bump: patch
+type: fix
+---
+
+Record one event for a bulk enqueue of Active Job jobs. Enqueueing jobs with
+`ActiveJob.perform_all_later` recorded an event for the batch and then another
+event for each job in it. A job that slices a large collection and enqueues it
+in batches was therefore reported with an event per job enqueued, where before
+version 4.9.0 it had one event per batch.
```
unflxw:
```suggestion
Record a single event for a bulk enqueue of Active Job jobs. Before this change, enqueueing jobs with
`ActiveJob.perform_all_later` would record an event for the batch and, if the
adapter was instrumented with AppSignal, another event for each job in it.
```

tombruijn (approved):

PR #1592, opened 2026-08-27
Report Delayed Job jobs that fail to load

unflxw:
The Delayed Job fix from #1590, forward-ported to the 5.x line.

### [Report Delayed Job jobs that fail to load](https://github.com/appsignal/appsignal-ruby/commit/f67a93aa5e7aba4c7f8a076169f97ba1a759f279)

The Delayed Job integration reads a job's payload after performing it,
to name the transaction and record the job's parameters. Delayed Job
raises when a job's handler no longer deserializes, and raises again on
every further read, so that read raises inside the `ensure` block. The
transaction is therefore never completed and the failure never reaches
AppSignal, even though the error is already set on it. The payload read
is now guarded, and a job whose payload cannot be read is named after
the class recorded in its raw handler, or after Delayed Job itself when
the handler does not record one either.

### [Test the Delayed Job payload fix in both modes](https://github.com/appsignal/appsignal-ruby/commit/0613744b129612132013e23e838e69815084745b)

The examples the fix arrived with only assert the agent-mode shape of
the report, because they came from a branch that has no collector mode.
Tag them and add the collector-mode half. The collector-mode examples
assert that the consumer span carries no `appsignal.ignore_subtrace`,
which is what says the collector will keep it: a span reaching the
exporter only means it was exported, and a flagged subtrace is exported
and then dropped.

### [Report Delayed Job jobs the worker cannot load](https://github.com/appsignal/appsignal-ruby/commit/02607a53a102fe48d691b341c6055039ca299fc8)

The plugin instruments a job from Delayed Job's `invoke_job` callback.
Delayed Job reads a job's payload before it invokes it, because it asks
the job for its own maximum run time, so a job whose payload will not
deserialize raises inside the worker and never reaches `invoke_job`.
The worker fails that job on its own and nothing is reported. The
`perform` callback runs for every job the worker takes, including that
one, so reading the payload there is what gets the failure reported.

The transaction stays on `invoke_job`. `Delayed::Worker#run` rescues
every exception a job raises, so a transaction opened around it would
never see the job's error, and error reporting would stop for every
failing job with nothing to show that it had.

### DIFF
--- a/lib/appsignal/integrations/delayed_job_plugin.rb
+++ b/lib/appsignal/integrations/delayed_job_plugin.rb
@@ -103,16 +103,26 @@ module Appsignal
           transaction.set_error(error)
           raise
         ensure
-          payload = job.payload_object
-          if payload.respond_to? :job_data
-            # ActiveJob
-            job_data = payload.job_data
-            transaction.set_action_if_nil("#{job_data["job_class"]}#perform")
-            transaction.add_function_parameters_if_nil(job_data.fetch("arguments", {}))
-          else
-            # Delayed Job
-            transaction.set_action_if_nil(action_name_from_payload(payload, job.name))
-            transaction.add_function_parameters_if_nil(extract_value(payload, :args, {}))
+          # Delayed Job raises when a job's handler will not deserialize, and it
+          # raises again on every further attempt to read it. Reading the payload
+          # here without a guard therefore replaces the error the job already
+          # failed with, and skips the rest of this block, so the transaction is
+          # never completed and the failure is never reported.
+          begin
+            payload = job.payload_object
+            if payload.respond_to? :job_data
+              # ActiveJob
+              job_data = payload.job_data
+              transaction.set_action_if_nil("#{job_data["job_class"]}#perform")
+              transaction.add_function_parameters_if_nil(job_data.fetch("arguments", {}))
+            else
+              # Delayed Job
+              transaction.set_action_if_nil(action_name_from_payload(payload, job.name))
+              transaction.add_function_parameters_if_nil(extract_value(payload, :args, {}))
+            end
+          rescue => error
+            warn_unreadable_payload_once(error)
+            transaction.set_action_if_nil(action_name_without_payload(job))
           end
 
           transaction.add_tags(
@@ -128,14 +138,67 @@ module Appsignal
         end
       end
 
+      # The name Delayed Job derives from the raw handler when the payload will
+      # not deserialize. It reads the class name out of the handler with a
+      # regular expression, which covers the job class a deploy removed. That
+      # expression raises for a handler it does not match, and it is skipped
+      # altogether when the payload raised something Delayed Job does not wrap,
+      # so fall back to a name that says the failure came from inside Delayed
+      # Job. The failure is then reported under a name that can be found, and
+      # every job this happens to is grouped together.
+      def self.action_name_without_payload(job)
+        with_perform_suffix(job.name)
+      rescue
+        "DelayedJobInternal"
+      end
+
+      # Guards the check-and-set below, so two threads that read an unreadable
+      # payload at the same time cannot both warn. A constant so it is created
+      # once at load time, because creating it lazily would race in turn.
+      WARN_ONCE_LOCK = Mutex.new
+
+      # A deploy that removes a job class leaves every job of that class unable
+      # to deserialize, so this can be reached once per job. Each of those jobs
+      # reports its own error to AppSignal, so the log only has to say once that
+      # it is happening.
+      def self.warn_unreadable_payload_once(error)
+        should_warn = WARN_ONCE_LOCK.synchronize do
+          next false if @warned_unreadable_payload
+
+          @warned_unreadable_payload = true
+        end
+        return unless should_warn
+
+        Appsignal.internal_logger.warn(
+          "Unable to read a Delayed Job job's payload: #{error.class}: " \
+            "#{error.message}. Jobs whose payload cannot be read are reported " \
+            "without parameters. They are named after the class in their raw " \
+            "handler, or DelayedJobInternal when that cannot be read either."
+        )
+      end
+
+      # @!visibility private
+      #
+      # Resets the warn-once state. Only used to keep test runs isolated.
+      def self.reset_unreadable_payload_warning!
+        WARN_ONCE_LOCK.synchronize { @warned_unreadable_payload = false }
+      end
+
       def self.action_name_from_payload(payload, default_name)
         # Attempt to find appsignal_name override
         class_and_method_name = extract_value(payload, :appsignal_name, nil)
         return class_and_method_name if class_and_method_name.is_a?(String)
-        return default_name if default_name.split("#").length == 2
-        return default_name if default_name.split(".").length == 2
 
-        "#{default_name}#perform"
+        with_perform_suffix(default_name)
+      end
+
+      # An action name is a class and a method. A name that already names both,
+      # separated either way, is left alone.
+      def self.with_perform_suffix(name)
+        return name if name.split("#").length == 2
+        return name if name.split(".").length == 2
+
+        "#{name}#perform"
       end
 
       # rubocop:disable Style/OptionalBooleanParameter
--- a/lib/appsignal/integrations/delayed_job_plugin.rb
+++ b/lib/appsignal/integrations/delayed_job_plugin.rb
@@ -11,6 +11,18 @@ module Appsignal
           enqueue_with_instrumentation(job, block)
         end
 
+        # Delayed Job asks a job for its own maximum run time before it invokes
+        # it, and answering that question reads the job's payload. A job whose
+        # payload will not deserialize therefore raises inside the worker and
+        # never reaches `invoke_job`, so the callback below does not run and the
+        # failure goes unreported. This callback sits one step further out, on
+        # the worker's whole run of the job, which is where such a job can still
+        # be seen.
+        lifecycle.around(:perform) do |worker, job, &block|
+          report_job_that_cannot_be_loaded(job)
+          block.call(worker, job)
+        end
+
         lifecycle.around(:invoke_job) do |job, &block|
           invoke_with_instrumentation(job, block)
         end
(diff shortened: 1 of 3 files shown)


spec/lib/appsignal/integrations/delayed_job_plugin_spec.rb
```diff
@@ -252,6 +256,115 @@ def perform
...
+              .to raise_error(Delayed::DeserializationError)
+
+            expect(root_span.kind).to eq(:consumer)
+            expect(root_span.attributes["appsignal.action_name"]).to be_nil
+            event = root_span.events.find { |e| e.name == "exception" }
+            expect(event.attributes["exception.type"])
+              .to eq("Delayed::DeserializationError")
+          end
+        end
+      end
```
unflxw, line 375: Curious about this: my understanding was that the agent would never flush a transaction with no action name. Would this also be true of transactions with errors in them?

In collector mode, we attempt to mimic my understanding above by discarding transactions for which an action name was never set.

If my understanding above is wrong, and the agent does flush transactions with no action name when they contain errors, then collector mode must mimic that, probably by forcing in some kind of `"[unnamed action]"` placeholder when an error has been set.

If my understanding was correct, then this does nothing in either agent mode and collector mode, and so a payload from which a class name cannot be deserialised goes unreported. It would be nice to fix that in some way while we're here -- perhaps we make up some communicative-of-the-scope-though-not-necessarily-correct name for the action, like "Delayed::Job#perform".

This comment may have implications for #1590 as well.

unflxw (commented):
Would be easier to review if it was split into two commits -- a cherry-pick of #1590, then the dual-mode changes stacked on top.

spec/lib/appsignal/integrations/delayed_job_plugin_spec.rb
unflxw: My understanding here was wrong. The agent does flush an errored transaction with no action, so collector mode now forces in `[unnamed action]`. The findings and the reasoning are in https://github.com/appsignal/appsignal-ruby/pull/1597#discussion_r3893171509.

The `Delayed::Job#perform` suggestion landed too, in #1590 and in this PR.

lib/appsignal/integrations/delayed_job_plugin.rb
```diff
@@ -103,16 +103,26 @@ def self.invoke_with_instrumentation(job, block)
...
-            # Delayed Job
-            transaction.set_action_if_nil(action_name_from_payload(payload, job.name))
-            transaction.add_function_parameters_if_nil(extract_value(payload, :args, {}))
+          # Delayed Job raises when a job's handler will not deserialize, and it
+          # raises again on every further attempt to read it. Reading the payload
+          # here without a guard therefore replaces the error the job already
+          # failed with, and skips the rest of this block, so the transaction is
+          # never completed and the failure is never reported.
+          begin
+            payload = job.payload_object
```
tombruijn, line 112: So it only fails with the deserialization error on calling `job.payload_object`, not just any other field on the `job` object, right?
Meaning it won't fail on lines 88 and 98 for example.

lib/appsignal/integrations/delayed_job_plugin.rb
```diff
@@ -128,14 +138,68 @@ def self.invoke_with_instrumentation(job, block)
...
+      # regular expression, which covers the job class a deploy removed. That
+      # expression raises for a handler it does not match, and it is skipped
+      # altogether when the payload raised something Delayed Job does not wrap,
+      # so name Delayed Job itself when there is nothing else to go on. The
+      # failure is then reported under a name that can be found, and every job
+      # this happens to is grouped together.
+      def self.action_name_without_payload(job)
+        with_perform_suffix(job.name)
+      rescue
+        "Delayed::Job#perform"
```
tombruijn, line 152: For Sidekiq I called this `SidekiqInternal` to try and communicate it was from a Sidekiq internal error. Something like that might be more clear than `Delayed::Job#perform`?

lib/appsignal/integrations/delayed_job_plugin.rb
unflxw: That's the idea, yes. I'll double-check.

lib/appsignal/integrations/delayed_job_plugin.rb
unflxw: This was an intentional decision. I did consider the Sidekiq precedent, but I thought that a name that indicated in some way what the action that we were attempting was (in this case, through some non-existing "Delayed::Job#perform" method) was preferable to a more generic "DelayedJobInternal" that didn't communicate what it was specifically in Delayed Job that triggered this.

It's in no way a strongly held opinion, I'll change it to "DelayedJobInternal".

lib/appsignal/integrations/delayed_job_plugin.rb
unflxw: Double-checked this against the Delayed Job source, from 3.0.5 up to 4.2.0. The answer to your question is yes, but the check turned up a bigger problem underneath it.

On the fields we read, you are right. `Delayed::Backend::Base` only reads the payload in `name`, `payload_object`, `invoke_job`, `hook`, `reschedule_at`, `max_attempts`, `max_run_time` and `destroy_failed_jobs?`. Everything this plugin reads off the job is a plain database column instead: `queue` on lines 88 and 98, and `id`, `queue`, `priority`, `attempts` and `run_at` in the tags below. None of those touch the handler, so none of them can raise the deserialization error. That has been true in every version since 3.0.5. I also confirmed it by running an unreadable job on 4.2.0. Every one of those fields returned its value, and `job.name` returned the class name it read out of the raw handler.

The bigger problem is that three of the methods that do read the payload raise the error, and Delayed Job calls one of them itself before it ever reaches us. `Delayed::Worker#run` passes `max_run_time(job)` to `Timeout.timeout` before it calls `job.invoke_job`. `Delayed::Worker#max_run_time` calls `job.max_run_time`, which reads the payload. So on a real worker the error is raised outside `invoke_job`, `Worker#run` catches it in its own `rescue DeserializationError`, and the `around(:invoke_job)` callback this plugin registers never runs at all. No transaction is created, and nothing is reported.

I confirmed that too. With the plugin registered, `Delayed::Worker#run` on an unreadable job never entered `invoke_with_instrumentation`. The specs here pass because `perform_job` calls `job.invoke_job` directly, and that is the one path that does reach the callback.

This affects Delayed Job 4.1.0 and up. Before 4.1.0, `Worker#run` used the class-level `Delayed::Worker.max_run_time` and never asked the job, so the error first surfaced inside `invoke_job`, where our callback already sits. 4.1.0 is from 2015, so in practice every user is on an affected version. It also means the changeset promises more than we deliver: a worker does not report these jobs today, with or without this change.

The guard itself is still worth keeping. An exception escaping this ensure block loses the transaction, and the naming path can raise on its own, because `job.name` falls back to `ParseObjectFromYaml.match(handler)[1]` and that raises `NoMethodError` for a handler the expression does not match. But the removed-class case the guard was written for needs a callback that sits further out.

Two candidates, both of which I checked do fire for such a job. `around(:perform)` wraps `Worker#run`, so it can read the payload before the worker does. `after(:failure)` is run by Delayed Job from that `rescue DeserializationError` branch, with `job.error` already set. I would rather settle which one owns the transaction in a follow-up than grow this pull request, and the same follow-up would apply to #1590.

unflxw: Two corrections and some more detail, after running this per version rather than reading the source alone.

First, I had the version boundary wrong. It is 4.0.4, not 4.1.0. `Delayed::Backend::Base#max_run_time` and the `max_run_time(job)` call in `Worker#run` both arrived in 4.0.4, and 4.1.0 only inherited them. I checked 4.0.0 through 4.0.6 one release at a time to place it. I also confirmed it by running an unreadable job on 4.0.4 and on 4.0.6: in both, the `around(:invoke_job)` callback never ran. So the affected range is wider than I said, and it starts in 2014.

Second, `around(:perform)` holds up across every version I could run. I ran the same worker-level reproduction on 4.0.4, 4.0.6, 4.1.0, 4.1.5, 4.1.11, 4.1.13 and 4.2.0. In all of them the `:perform` callback exists, is invoked, wraps `Worker#run` in the same place, and can read the payload before the worker does. `after(:failure)` fires in all of them too. The lifecycle event list is byte-identical all the way back to 3.0.5, where `:perform` is already declared as `[:worker, :job]`.

The versions I could not run are 3.0.5 through 4.0.3, and not for a reason that has anything to do with us. 3.0.5 requires `active_support/basic_object`, which Rails removed in 4.1. Everything up to 4.0.3 calls `alias_method_chain`, which Rails removed in 5.1. Neither of those Active Support versions runs on any Ruby we test, so those Delayed Job versions cannot be loaded on a supported Ruby at all. For those I am going on source reading, which says they are the unaffected ones, because their `Worker#run` uses the class-level `Delayed::Worker.max_run_time` and never asks the job.

One thing to be careful about when we do this. `:perform` wraps a wider span of work than `:invoke_job` does. I traced the order for a job that loads fine, and it is `:perform` in, payload read, `:invoke_job` in, `:invoke_job` out, `job.destroy`, `:perform` out. So `:perform` also covers the payload deserialization, the `Timeout` setup and the database write that deletes the finished job. More importantly, `Worker#run` rescues every exception itself, so nothing reaches an `around(:perform)` block. Our `set_error` call depends on the job's exception reaching us, and at `:perform` it never would.

That is why I would keep `around(:invoke_job)` as the instrumentation that owns the transaction, and add a narrow `around(:perform)` that only handles the case where the payload cannot be read. Jobs that load fine then keep exactly the timings they have now.

unflxw: Double-checked, and the answer to your question is yes. Every field the plugin reads off the job is a plain database column, including the `queue` read on lines 88 and 98, so none of them can raise the deserialization error. That has been true in every Delayed Job version since 3.0.5.

The check did turn up something bigger. `Delayed::Worker#run` reads the payload itself, through `max_run_time(job)`, before it calls `invoke_job`. On a real worker the error was therefore raised outside `invoke_job`, the `around(:invoke_job)` callback never ran, and nothing was reported at all. That starts in 4.0.4. The findings and the measurements are in [my earlier comment](https://github.com/appsignal/appsignal-ruby/pull/1592#discussion_r3948641198) and [the correction below it](https://github.com/appsignal/appsignal-ruby/pull/1592#discussion_r3948796774).

That is fixed in the third commit. A narrow `around(:perform)` callback reports a job whose payload cannot be read, and the transaction stays on `invoke_job`, so an ordinary job's timings and its reported error are unchanged. The same change is on #1590.

lib/appsignal/integrations/delayed_job_plugin.rb
unflxw: The rename landed. The fallback name is `DelayedJobInternal` now, here and in #1590.

PR #1595, opened 2026-08-27
Turn the agent's OpenTelemetry listener off

unflxw:
The gem sends no OpenTelemetry data to the agent in either mode, so the
agent has no reason to listen for it. It does not listen today either,
because the setting defaults to off, but that default lives in the
agent and a value set in the environment reaches it before the gem
writes anything.

Writing the setting out makes the gem decide it. The port it binds is
also the collector's default port, so an agent listening on it next to
a collector on the same host is a conflict waiting to happen.

AppSignal for Python does the same as part of appsignal/appsignal-python#280.

[skip changeset]


unflxw:
Merging 'cause it has no effect, just making the decision explicit.

PR #1596, opened 2026-08-27
Send collector data through the CA file and proxy

unflxw:
The OTLP exporters are built with an endpoint and nothing else, so the
`ca_file_path` and `http_proxy` options do not reach the data sent in
collector mode. They do reach the data the agent sends, so a proxy or a
private certificate authority works in one mode and not the other.

The certificate file is a keyword argument the exporters accept. The
proxy is not, and they expose no way to reach the connection they send
through, so an exporter that needs a proxy is a subclass that applies
it to the connection its own `http_connection` returns.

That method is private, and the version range in `REQUIRED_GEMS` allows
a version that changes it to reach us. The override forwards whatever it
is called with, so an argument added to it does not raise on the way
through. It also records whether it managed to configure a connection,
so a version that no longer builds one there is reported in the log
instead of quietly sending around the proxy. A spec asserts the proxy
reached all three exporters, and an integration spec runs an application
against a mock proxy to check that the requests really go through one.

The same change for AppSignal for Python is in appsignal/appsignal-python#281.

### DIFF
--- a/lib/appsignal/opentelemetry/proxied_exporter.rb
+++ b/lib/appsignal/opentelemetry/proxied_exporter.rb
@@ -1,0 +1,83 @@
+# frozen_string_literal: true
+
+require "uri"
+
+module Appsignal
+  module OpenTelemetry
+    # @!visibility private
+    #
+    # Routes an OTLP exporter's requests through the proxy in the
+    # `http_proxy` config option.
+    #
+    # The exporters accept no proxy and expose no way to reach the connection
+    # they build. They do build it in one method, `http_connection`, which
+    # this module overrides to configure the connection it returns.
+    #
+    # This module is included into a subclass rather than prepended onto the
+    # exporter itself, so that an application using the OpenTelemetry gems for
+    # its own exporting is unaffected.
+    #
+    # That method is not part of the exporters' public API, so a later version
+    # can rename it, stop calling it, or return something other than a
+    # `Net::HTTP` from it. This module does not try to predict which of those
+    # happened by inspecting the method. It records whether it managed to
+    # configure a connection, and {#initialize} reports it when it did not, so
+    # a version we cannot proxy through is visible in the log rather than
+    # silently sending around the proxy.
+    module ProxiedExporter
+      def initialize(appsignal_http_proxy:, **kwargs)
+        @appsignal_http_proxy = appsignal_http_proxy
+        @appsignal_proxy_applied = false
+
+        # The exporters build their connection while they initialize, so by
+        # the time this returns the override below has either run or never
+        # will.
+        super(**kwargs)
+
+        return if @appsignal_proxy_applied
+
+        Appsignal.internal_logger.error(
+          "Not sending #{self.class.superclass} data through the proxy in " \
+            "the `http_proxy` option: this version of the OpenTelemetry " \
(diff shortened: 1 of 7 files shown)


unflxw:
Leaving this one up for human review (as opposed to the Python one, which I merged without review because it's quite straight-forward) because it subclasses OpenTelemetry's exporters to override a private method, which is dicey.

I do think it's worth doing it -- the alternative is telling customers that our config option doesn't actually work in collector mode, and they need to set the `http_proxy` environment variable instead, and warn them that it would likely also proxy all other traffic sent by their application, not just AppSignal's, which may not be what they want at all, oh and also _technically_ supporting `http_proxy` is _itself_ an implementation detail of the OpenTelemetry's exporters (it happens because `Net::HTTP` does it) so it's not much less liable to change between releases than the method we're overriding anyway. 🫠

(The risk of it being changed is why there's an integration test for it -- CI should always run against the top of the OpenTelemetry dependency version ranges, so if it breaks at least we'll find out)

lib/appsignal/opentelemetry/proxied_exporter.rb
```diff
@@ -0,0 +1,43 @@
...
+  module OpenTelemetry
+    # @!visibility private
+    #
+    # Routes an OTLP exporter's requests through the proxy in the
+    # `http_proxy` config option.
+    #
+    # The exporters accept no proxy and expose no way to reach the connection
+    # they build. They do build it in one private method, `http_connection`,
+    # which this module overrides to configure the connection it returns.
+    #
```
tombruijn, line 15: This sounds a bit fragile. If they rename the method or change it, it breaks for us?

unflxw: Yes, it is. If they rename or delete the method, it breaks in that it no longer sends data via the proxy, attempting to send it through the network instead -- same as if we didn't override it in this way.

If they change the method's signature or the expectations on how it is used, that could break in the same way or crash the application. We can attempt to prevent this by only overriding the method if the signature is the same. I'll look into doing that here, and also in #1591, which also instruments a somewhat-private method.

In either case, we have an integration test that would catch it after the fact, so we wouldn't be relying on our customers to find out.

We could also tighten the version bounds for the OpenTelemetry dependencies to ensure we don't accidentally allow a version with different expectations of that method to be used, but I'd rather avoid that.

unflxw: What was done to address this here:
- Expand the signature from `*args` to `*args, **kwargs` -- this ensures the override cannot fail when it calls `super`.
- Keep track of whether the proxy was indeed applied in the initialiser, and log something if it wasn't.
- Have a unit test that checks for the variable that keeps track specifically, separately from the integration tests that check that it does talk through the proxy. That way the failure is named.

It was addressed differently in #1591. There we check for the specific shape of the method and refuse to instrument bulk enqueues if it changes: https://github.com/appsignal/appsignal-ruby/pull/1591/changes/c5ca75e9fac8ff4e9769891144cf6a33ba229a14#diff-4392a75b3f401aced5408c5878860c62fc89dc7db6e36c0b30679aee46991f36R24-R75

tombruijn (approved):

PR #1597, opened 2026-08-27
Report errored transactions with no action

unflxw:
Collector mode flags a transaction that never set an action with
`appsignal.ignore_subtrace`, so the collector drops it, on the grounds
that agent mode does not report one either. That is only half true. The
agent drops a transaction when it has neither an action nor an error, to
leave out noise like asset requests in development, and an error exempts
it. So a failure that AppSignal could not name was reported in agent
mode and thrown away in collector mode. Draw the line where the agent
draws it, and keep flagging a transaction that recorded neither.

Reporting one needs a name to report it under, and an action left unset
does not stay empty. The collector fills an empty action in from the
span name, which here is the placeholder the transaction span opened
with, and that reads as though it were the application's own action. So
these are named `[unnamed action]`, which says what it is. The aggregate
metrics were gated on an action having been set, standing in for the
transaction being reported at all, and they gate on that directly now.
The name is settled before they run, because they tag by it.

[skip changeset]

### DIFF
--- a/lib/appsignal/transaction/opentelemetry_backend.rb
+++ b/lib/appsignal/transaction/opentelemetry_backend.rb
@@ -14,6 +14,12 @@ module Appsignal
     class OpenTelemetryBackend < BaseBackend
       TRACER_NAME = "appsignal-ruby"
 
+      # The action name given to a transaction that recorded an error without
+      # ever setting one. Bracketed so it reads as a marker rather than as a
+      # class the application defines, and so it sorts apart from real action
+      # names.
+      UNNAMED_ACTION = "[unnamed action]"
+
       # Guards the process-wide warn-once state, which transactions touch
       # concurrently on threaded servers. A constant so it is created once at
       # load time rather than lazily (which would race).
@@ -125,6 +131,7 @@ module Appsignal
         @start_time = Time.now
         @action_set = false
         @action = nil
+        @error_set = false
         @allocation_start = current_allocation_count
         @root_child_allocation_count = 0
 
@@ -317,6 +324,7 @@ module Appsignal
       # become their own incident. Each cause carries only the part of its
       # backtrace that is not shared (see `trim_shared_tail`).
       def set_error(class_name, message, backtrace, causes, _root_cause_missing)
+        @error_set = true
         span = current_span
         error_lines = Array(backtrace)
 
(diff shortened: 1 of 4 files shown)


lib/appsignal/transaction/opentelemetry_backend.rb
```diff
@@ -450,13 +452,19 @@ def teardown
...
+      # collector to drop instead, the same way `discard` does. The flag has to be
+      # set before `teardown` finishes the span, because attributes set on an
+      # ended span are dropped.
+      #
+      # A transaction that recorded an error is reported whether or not it has an
+      # action, because the failure is worth reporting even when we cannot say
+      # what failed. The agent draws the line in the same place: it drops a
+      # transaction only when it has neither an action nor an error, to leave out
+      # noise like asset requests in development.
       def ignore_subtrace_without_action
```
unflxw: Nit: `ignore_subtrace_without_action` is not a good name anymore. Probably a good reason, alongside with what's below, to re-structure this a bit.

More important: we should find out what the collector does when there's no explicit action name set. I believe what it would do is use the root span's name instead, we should verify that. We should compare that to what the agent data pipeline does (the agent, then the processor, then how the server displays it -- does it just leave it empty throughout?) and then decide what we want to do. Are we okay with using the root span name, if that's what the collector currently does? If not, would we rather match the agent's behaviour exactly by forcing through an empty action name? Or would we prefer to put some explicit `[unnamed action]`-like placeholder?

unflxw: Looked into this.

The collector does fall back to the span name. `complete_root_span` sets the action from `action_name_for_root`, which prefers `appsignal.action_name` and uses the span name when that is empty. For our spans the span name is the placeholder, so an unnamed transaction would have been reported as `appsignal.transaction background_job`.

The agent leaves the action empty, and nothing downstream fills it in.

The server does not treat empty as deliberate. The main error page renders the action link only when there is an action, so it shows nothing at all. Three other places invent their own placeholder, and none of them agree: `<Unknown action name>`, `(unknown action)`, `Unknown action`.

So neither existing behaviour is a considered answer. This now sets `[unnamed action]`. It reads as a marker rather than as a class we defined, it is one string instead of three, and it passes the `action_name != ''` filter the trace views use.

The marker is only set when there is an error to report. Setting an action unconditionally would defeat the drop rule and start reporting the noise that rule exists to leave out.

On the naming: `ignore_subtrace_without_action` is gone. There is a `should_report?` predicate now, which both metric gates use, and `resolve_missing_action` decides between naming and flagging.

The collector's span-name fallback is still wrong for our spans either way. I will raise that on the collector separately.

tombruijn (approved):

PR #1609, opened 2026-09-08
Separate request headers from the Rack environment

unflxw:
The `request_headers` configuration option is an allowlist of Rack environment key
names. Some of those names are HTTP request headers, and some are CGI variables
that are not headers at all. Collector mode needs the two kinds apart, because
OpenTelemetry reports a header as an `http.request.header.*` attribute and has
nowhere to put the rest.

Since collector mode shipped, no application using it has reported a single
request header, because the gem sends the Rack spellings and the collector
matches them against the OpenTelemetry ones. Webmachine applications have
reported none in either mode, for the mirror image of the same reason.

This splits the two kinds at the source, so our own instrumentation says which
kind it is reporting rather than leaving AppSignal to guess. The same rule then
gets applied to the parameters options, which had the same problem. One rule now
covers every option that predates collector mode: the agent honours it, the
collector ignores it, and collector mode reads it only to work out a value for
the options that replace it.

## Related pull requests

- [appsignal-docs#245](https://github.com/appsignal/appsignal-docs/pull/245) documents the new options and corrects what the existing option pages say about collector mode.
- [diagnose_tests#158](https://github.com/appsignal/diagnose_tests/pull/158) updates the shared diagnose test expectations for the new options and for the derived source. The submodule pointer here already points at it, so that has to merge first.
- [appsignal-collector#437](https://github.com/appsignal/appsignal-collector/pull/437) makes the collector match a header name whichever spelling it arrives in, which is what lets the normalized names this sends reach a trace.

### [Do not set an option a config block only reads](https://github.com/appsignal/appsignal-ruby/commit/c6abb7a0b0cc6e5f78767f47f4f2ed31722bb7b5)

The configuration DSL copies an option's current value when a block
reads it, so that appending to it works, and records that copy as a
value the block set. An option AppSignal derives from another then keeps
whatever a block read, because a value the block set is one deriving
leaves alone. Record an option only when the block assigns it or changes
its value.

### [Map a request header's two names in one place](https://github.com/appsignal/appsignal-ruby/commit/6e64948e1a73672e0ca6f29570af8d6e6cdde440)

The OpenTelemetry backend decides which Rack environment keys hold a
request header, and converts those keys to the names OpenTelemetry uses.
The extension backend needs the same rules in reverse, to report Rack
names for headers the instrumentation hands over under their
OpenTelemetry names. Move both directions and the split to
`Appsignal::Utils::RequestHeaders`, so one place decides which keys are
headers.

### [Add the collector-mode header allowlists](https://github.com/appsignal/appsignal-ruby/commit/1cc97e68f2a22f4add07cb1fb54ae7361d71d74f)

Collector mode sends `request_headers` to the collector, which uses it
as an allowlist over the `http.request.header.*` attribute names. Those
names are the ones OpenTelemetry uses, while `request_headers` names
Rack environment keys, so `accept` never matched `HTTP_ACCEPT` and the
collector dropped every header. Add `keep_request_headers` and
`keep_request_environment` to name the two kinds separately, send the
first under the attribute the collector already reads, and derive both
from `request_headers`.

AppSignal tells a value it worked out apart from one the application
set. Worked-out values go in a config source of their own, ranked below
every other source, so `appsignal diagnose` names that source, an option
the application set takes precedence over it, and the warnings about
options that only apply in the other mode fire only for options the
application set.

### [Send every empty list to the collector](https://github.com/appsignal/appsignal-ruby/commit/fa98aa15eded8b3b1a6db28f106ea948e02d49ee)

`build_resource` drops an attribute whose value is empty, which is how
an unset string option arrives. An empty allowlist means "keep nothing",
which no absent attribute can say, so an empty `response_headers`
reached the collector as no attribute and kept every header. Drop empty
strings only. An empty denylist filters nothing whether or not the
attribute arrives.

### [Show derived options in the diagnose report](https://github.com/appsignal/appsignal-ruby/commit/0a3317c040e17a7da5c6cd06ecd380166c87315b)

The diagnose command loads the config and prints every option with the
source its value came from. `Appsignal.start` resolves the options
AppSignal derives from other options, and the command skips it, so a
derived option printed as not configured. Resolve them in the command as
well.

### [Store request headers apart from the environment](https://github.com/appsignal/appsignal-ruby/commit/b3426fbb273ad309f0d7994fddac8e509c269d70)

The transaction keeps everything given to `add_headers` in one bucket,
and the OpenTelemetry backend decides at sample time which of those keys
hold a request header. A Rack environment mixes headers with CGI
variables, and nothing in the bucket says which kind a value is. Give
the transaction a channel per kind, mapped to storage buckets by the
backend the way the params channels are. Agent mode maps both to the one
environment bucket it has a slot for, and converts a header's
OpenTelemetry name back to its Rack name. Collector mode keeps them
apart, each filtered by its own allowlist.

### [Guard every _if_nil setter on its own channel](https://github.com/appsignal/appsignal-ruby/commit/4a2356743063a1336dec2878fcf9e7278702c3f8)

The `_if_nil` params setters ask whether the storage bucket their
channel maps to holds a value, and agent mode maps every params channel
to one bucket. Setting the request payload there stops an integration
from setting the function parameters, while collector mode keeps both,
because it gives each channel a bucket. Track what has been set per
channel, the way the header channels do, so both modes report the same
values. `add_params` names no channel, so it records as the request
payload, which is the channel it maps to in collector mode.

### [Report headers and environment values separately](https://github.com/appsignal/appsignal-ruby/commit/93a301e4e4cd9d03e97dc68629dfcc07a4097050)

`Appsignal.add_headers` takes a Rack environment, which mixes request
headers with CGI variables. Collector mode reports the two kinds under
different span attributes. Add `add_request_headers` and
`add_request_environment`, so a caller says which kind it is reporting.

### [Say what our instrumentation is reporting](https://github.com/appsignal/appsignal-ruby/commit/838f07834501d46dd2a467ac45ad4e98fe7b5f61)

The Rack, Action Cable and Webmachine instrumentation hands
`add_headers` whatever the request object gives it, leaving AppSignal to
work out which of those values are headers. Each integration knows the
answer already: a Rack environment mixes the two kinds, and Webmachine's
headers are only headers. Call the two channels instead, so each
integration says which kind it reports. Webmachine's headers are named
the way OpenTelemetry names them, so this is also what gets them past a
Rack-named allowlist.

### [Deprecate request_headers in collector mode](https://github.com/appsignal/appsignal-ruby/commit/c4e2cb2b8ff749e22483f196d0d0332fc5c6e51d)

Collector mode reads `request_headers` to derive `keep_request_headers`
and `keep_request_environment`, and those two say which names they
expect. Warn at startup that the option is deprecated in collector mode,
and name the value to set for each option that replaces it, which
AppSignal has already worked out in order to fall back to it.

### [Filter each kind of params by its own option](https://github.com/appsignal/appsignal-ruby/commit/f4a0ac98b55f32a455ea22881afd6fc384e0d4ef)

Collector mode keeps the request payload, the function parameters and
the query parameters in an attribute each, with an option per kind to
filter and to suppress them. The gem applies `filter_parameters` and
`send_params` to every bucket, so an option set for one kind does
nothing until the data reaches the collector, and `filter_parameters`
filters a job's arguments as well as a request's parameters. Give each
bucket the pair of options that names it, the way each header bucket has
its allowlist, and derive the six from the two. Agent mode has one
bucket, and keeps the two options it has always used.

### [Stop warning that filter_metadata is ignored](https://github.com/appsignal/appsignal-ruby/commit/c9b13bab616b0ad09f2b22a0f6f388cee18adb28)

`Transaction#set_metadata` applies `filter_metadata` before the value
reaches either backend, so it filters metadata in both modes. Drop it
from the list of options the collector ignores, which empties the list,
and drop the list and its warning.

### [Deprecate add_headers in the documentation](https://github.com/appsignal/appsignal-ruby/commit/635522ce8a852fe3365aed9f7a0df0f46a27d73d)

`add_headers` takes request headers and Rack environment values in one
hash, and `add_request_headers` and `add_request_environment` take them
separately. Mark it deprecated in its documentation, pointing at both.
The method is on the request path, so the deprecation stays out of the
logs.

### [Name a header the way the convention does](https://github.com/appsignal/appsignal-ruby/commit/033b8449ab717568207eb2107ab91883263ad39a)

The OpenTelemetry semantic convention names a header's span attribute
after the header, lowercased with its dashes kept, such as
`accept-encoding`. Some SDKs write the older spelling with underscores,
and an application writes the name in its own case. The collector
compares headers by exact match against its `request_headers` allowlist,
so a name in another spelling keeps that header out.

Normalize a header's name to the convention's spelling before comparing
or sending it, both for the allowlists and for the span attributes that
carry header names.

### [Correct an array option set to a YAML null](https://github.com/appsignal/appsignal-ruby/commit/344038a46d1710cb900cc536eb345f3f2cd04605)

An array option written as a null in the config file holds `nil`. The
options that reach the agent are joined into a string when AppSignal
starts, and `filter_metadata` and `filter_session_data` are searched
while a transaction is sampled, so both raise a `NoMethodError`. Correct
such an option to the empty list it means, with the other overrides,
after the values derived from other options have been merged.

### [Say what to use instead in agent mode](https://github.com/appsignal/appsignal-ruby/commit/13914b927299a22de3a2f57f023cf27a8a6bb325)

Setting a collector-only option while the agent is in use warns that the
option is ignored, and repeats the way out of the warning on every line
of it. Name the agent-mode option that does the same thing, which eight
of the eleven options have, and say how to use the collector once at the
end.

### [Point the diagnose tests at the new options](https://github.com/appsignal/appsignal-ruby/commit/4d34d4578d75b886cf5cfb7060018c112eee8242)

The diagnose integration tests live in a repository shared with the
other language integrations, and assert the full list of options a
report prints and the sources it loads them from. Two options are new
here, and derived values come from a source of their own, so the
expectations there move with them.

### [Default the send options to sending](https://github.com/appsignal/appsignal-ruby/commit/5f971c508d93d8372b231bcf1d3ddaf36501ae90)

`send_request_payload`, `send_request_query_parameters` and
`send_function_parameters` have no default, while deriving them from the
default `send_params` gives `true`. A diagnose report of an application
that configured none of them prints `nil` for the three and `true` for
the option they are derived from. Give them the value the derivation
would give, and hold every pair in the table to that rule in a spec.

The resource sends the three attributes now rather than leaving them
out. The collector suppresses a kind of parameter only on an explicit
`false`, so it reads an attribute that says `true` and an absent one the
same way.

[skip changeset]


lib/appsignal/opentelemetry.rb
```diff
@@ -255,7 +262,17 @@ def build_resource(config)
...
+        # default does not, so it is dropped. The two allowlists are the
+        # exception: the collector reads an absent one as "keep everything" and
+        # an empty one as "keep nothing", so an empty allowlist has to be sent.
+        # An unset allowlist holds `nil` and is dropped like anything else.
+        attrs.reject! do |key, value|
+          next true if value.nil?
+          next false if ALLOWLIST_ATTRIBUTES.include?(key)
+
+          value.respond_to?(:empty?) && value.empty?
+        end
```
unflxw: Wondering if we could avoid checking `empty?` generally, and check it only for strings, where it's useful, and not have this added complexity of treating array attributes differently depending on whether they're allowlists or denylists.

.changesets/allow-an-empty-response-headers-option.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: patch
+type: fix
+---
+
+Setting the `response_headers` configuration option to an empty list now reports no response headers in collector mode. Before this change an empty list had no effect, so every response header captured by the application's own OpenTelemetry instrumentation was reported anyway.
+
+Leaving the option unset is unchanged and still reports every captured response header.
```
unflxw: Not sure if that's the right default to have.

I'm blanking right now on whether this config option existed in agent mode at all. If it did, then whatever was the default behaviour in agent mode should be the default behaviour now. If it didn't, then I'd suggest undoing the change in `config.rb:147` and reporting no response headers by default as a preferable default behaviour.

.changesets/report-request-headers-in-collector-mode.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: patch
+type: fix
+---
+
+Report request headers again in collector mode. Since version 5.0.0.rc.1, no application in collector mode reported any request header. The `request_headers` configuration option lists Rack environment keys, such as `HTTP_ACCEPT`, and the collector matched those against the OpenTelemetry names for the same headers, such as `accept`. Nothing matched, so every header was dropped.
```
unflxw: We don't need to specify "since version 5.0.0.rc.1", as that's the only released version with collector mode.

lib/appsignal/config.rb
```diff
@@ -916,9 +922,41 @@ def determine_overrides
...
+    # header. Collector mode needs the headers under the names OpenTelemetry
+    # uses and the rest under their Rack names, so an application that only
+    # ever set `request_headers` keeps reporting the same values once it moves
+    # to collector mode.
+    #
+    # The environment half leaves out the keys the instrumentation already
+    # describes with a semantic convention attribute. An application that asks
+    # for one of them by setting `keep_request_environment` itself still gets
+    # it.
+    def derive_keep_options(config)
```
unflxw: Let's not call them "keep options" -- they do happen to share the `keep_` prefix today, but what they share in terms of derivation rules is that they're derived from the old `request_headers`, so what would be semantically meaningful is something like `derive_request_headers_options`.

.changesets/show-derived-options-in-the-diagnose-report.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: change
+---
+
+The `appsignal diagnose` report now shows the value of configuration options that AppSignal works out from other options, and names `override` as where the value came from.
```
unflxw, line 6: If we want to rename this to "derived" rather than "override", now would be a good time to do that.

(If we do, let's propagate the change through internal naming, e.g. `apply_derived` rather than `apply_overrides`)

spec/lib/appsignal/cli/diagnose_spec.rb
```diff
@@ -907,6 +907,34 @@ def dont_accept_prompt_to_send_diagnostics_report
...
+            ENV["APPSIGNAL_COLLECTOR_ENDPOINT"] = "http://collector.example.test:4318"
+            run
+          end
+
+          it "outputs the derived value and names the override as its source" do
+            expect(output).to include(
+              "  keep_request_environment: []\n" \
+                "    Sources:\n" \
+                "      default:  nil\n" \
+                "      override: []\n"
```
unflxw: Wondering if the default value for `keep_request_environment` should be `[]` instead of `nil`, and similarly the default value for `keep_request_headers` should be the derived list. They still can get overriden by `request_headers`, but only when `request_headers` has been modified and they haven't. This would make the diagnose less noisy when reporting on default configuration.

lib/appsignal/transaction/extension_backend.rb
```diff
@@ -96,6 +96,34 @@ def params_mapping
...
+      # That has to happen before the merge, because once the two channels
+      # share a bucket their values cannot be told apart.
+      HEADERS_MAPPING = {
+        :request_headers => [
+          :environment,
+          lambda do |headers|
+            headers.to_h do |name, value|
+              [Appsignal::Utils::RequestHeaders.rack_name(name), value]
+            end
+          end
```
unflxw: I think the transformation lambda would be more readable if it operated in each key-value pair, rather than in the whole `headers` -- that is, here we do `:request_headers => [:environment, lambda do |name, value| [BlahBlah.rack_name(name), value]]`, and when the lambda is called, instead of passing `headers` to it, we pass it as `headers.to_h(&transformation_lambda)`

.changesets/add-request-header-and-environment-helpers.md
```diff
@@ -0,0 +1,10 @@
+---
+bump: patch
+type: add
+---
+
+Add the `Appsignal.add_request_headers` and `Appsignal.add_request_environment` helpers, and their `set_` aliases.
```
unflxw: Do not add `set_` aliases for new helpers -- those only exist on other helpers for backwards compat.

.changesets/add-request-header-and-environment-helpers.md
```diff
@@ -0,0 +1,10 @@
...
+
+Use `add_request_headers` to report request headers, naming each header the way OpenTelemetry names it, in lowercase and with dashes, such as `accept`. Use `add_request_environment` to report the values a Rack environment holds that are not request headers, naming each one the way Rack names it, such as `REMOTE_ADDR`.
+
+Together they replace `Appsignal.add_headers`, which takes both kinds at once and has to work out which is which. `add_headers` still works and reports the same values it did before.
```
unflxw: Note here that `add_headers` is deprecated.

lib/appsignal/helpers/instrumentation.rb
```diff
@@ -874,13 +874,19 @@ def add_session_data(session_data = nil, &block)
       #   # The request headers will include:
       #   # { "PATH_INFO" => "/some-path", "HTTP_USER_AGENT" => "Firefox" }
       #
+      # @deprecated Use {#add_request_headers} for request headers and
+      #   {#add_request_environment} for the values a Rack environment holds
+      #   that are not request headers. This method takes both kinds at once,
+      #   so it has to work out which of them each value is.
```
unflxw: This commit should be squashed into the previous "deprecate add_headers" commit.

.changesets/stop-warning-that-filter-metadata-is-ignored.md
```diff
@@ -0,0 +1,6 @@
+---
+bump: patch
+type: fix
+---
+
+Stop warning that the `filter_metadata` configuration option is ignored in collector mode. It is not ignored. AppSignal filters the metadata it collects itself before reporting it, in collector mode as well as when the agent is in use.
```
unflxw:
```suggestion
Remove an incorrect warning about the `filter_metadata` configuration option being ignored in collector mode.
```

.changesets/deprecate-the-params-options-in-collector-mode.md
```diff
@@ -3,8 +3,8 @@ bump: patch
 type: change
 ---
 
-The `filter_parameters` and `send_params` configuration options are deprecated in collector mode. Use `filter_request_payload`, `filter_function_parameters` and `filter_request_query_parameters` to filter parameters, and `send_request_payload`, `send_request_query_parameters` and `send_function_parameters` to choose which parameters to report. Each of those covers one kind of parameter, where the deprecated options cover all of them at once.
+The `filter_parameters` and `send_params` configuration options are deprecated in collector mode. Each covers every kind of parameter at once, so neither can say anything about one kind on its own. Use `filter_request_payload`, `filter_function_parameters` and `filter_request_query_parameters` to filter parameters, and `send_request_payload`, `send_request_query_parameters` and `send_function_parameters` to choose which kinds to report.
```
unflxw:
```suggestion
The `filter_parameters` and `send_params` configuration options are deprecated in collector mode. Use `filter_request_payload`, `filter_function_parameters` and `filter_request_query_parameters` to filter different kinds of parameters, and `send_request_payload`, `send_request_query_parameters` and `send_function_parameters` to choose which kinds of parameters to report.
```

.changesets/deprecate-the-params-options-in-collector-mode.md
```diff
@@ -3,8 +3,8 @@ bump: patch
...
 
-Until they are set, their values are worked out from the deprecated options, so nothing changes for an application that only set those. AppSignal logs which values to set to keep reporting what it reports now.
+In collector mode, each kind of parameter is now filtered and reported according to its own options, and the two deprecated options have no other effect. Until the newer options are set, their values are worked out from the deprecated ones, so nothing changes for an application that only set those. AppSignal logs which values to set to keep reporting what it reports now.
```
unflxw:
```suggestion
In collector mode, each kind of parameter is now filtered and reported according to its own options, and the deprecated options' values are only used to fill in values when the new options are unset.
```

.changesets/deprecate-the-params-options-in-collector-mode.md
```diff
@@ -3,8 +3,8 @@ bump: patch
...
 
-Setting a deprecated option no longer logs that it is ignored in collector mode. It was never ignored: AppSignal applied it to the parameters it collected itself, and it still does. What it could not do is filter the parameters that an application's own OpenTelemetry instrumentation reports, which the newer options can.
+Nothing changes when the agent is in use. There, `filter_parameters` and `send_params` still apply to every kind of parameter, and the newer options still do nothing.
```
unflxw:
```suggestion
In agent mode, `filter_parameters` and `send_params` still apply to every kind of parameter, and the new options have no effect.
```

lib/appsignal/config.rb
```diff
@@ -915,9 +922,41 @@ def determine_overrides
...
+      request_headers = config_hash[:request_headers] || []
+
+      if config_hash[:keep_request_headers].nil?
+        config[:keep_request_headers] = request_headers.filter_map do |key|
+          Appsignal::Utils::RequestHeaders.header_name(key)
+        end
+      end
+
+      return unless config_hash[:keep_request_environment].nil?
```
unflxw: We check `keep_request_headers` and `keep_request_environment` against `nil`, but what we should actually be checking is "did any source other than `default` or `system` set these". This is likely true of all other overrides, and likely simplifies our whole `nil`-vs-`[]`-to-mean-unset problem space.

.changesets/show-derived-options-in-the-diagnose-report.md
unflxw: Ended up adding "derived" alongside "override". They serve different purposes -- "derived" provides user-overridable defaults based on other config options, while "override" provides final decisions on what a config option must be regardless of the user's wishes.

bjacquet (approved):

PR #1610, opened 2026-09-08
Support Grape 4

### LINKED ISSUE #1606: Support Grape 4
`Appsignal::Rack::GrapeMiddleware` builds the transaction action from the Grape endpoint's options Hash. It reads `options[:method]` for the HTTP method, `options[:path]` for the path and `options[:for]` for the API class:

```ruby
options = endpoint.options
request_method = options[:method].first.to_s.upcase
klass = options[:for]
path = options[:path].first.to_s
```

grape 4.0.0 moved all three of those off that Hash. They are now keyword arguments on `Grape::Endpoint#initialize`, which keeps them in a `Grape::Endpoint::Options` value object. [ruby-grape/grape#2775](https://github.com/ruby-grape/grape/pull/2775) moved `method` to `http_methods:`, [ruby-grape/grape#2776](https://github.com/ruby-grape/grape/pull/2776) moved `path`, and [ruby-grape/grape#2778](https://github.com/ruby-grape/grape/pull/2778) renamed `for:` to `api:`.

The `options` reader still exists and still returns a Hash, so the middleware's `unless endpoint&.options` guard still passes. The three lookups then return `nil`, and the first one raises `NoMethodError` on `nil.first`. That fails all seven examples in `spec/lib/appsignal/rack/grape_middleware_spec.rb`. It only fails on Ruby 3.3 and newer, because grape 4 requires that, so the older Rubies in the build matrix resolve grape 3 and keep passing.

The replacement is not a straight substitution. The value object is exposed as `Grape::Endpoint#config`, but that reader is `protected`, so it is not meant for us to call. Working this out is the main task here. `Grape::Router::Route` carries `request_method` and the path, and is reachable through `Grape::Endpoint#routes`, which looks like the most likely source. `endpoint.namespace` is unchanged.

CI is pinned to grape 3 as a workaround, and the gemset is renamed from `grape` to `grape-3`. That is #1607 on `main` and #1608 on `4.x-stable`. The `3.x-stable` line needs nothing, because its gemset already pins `grape` to `< 3`.

Supporting grape 4 means adding a `grape-4` gemset next to `grape-3`, so that both majors stay covered.

unflxw:
Fixes https://github.com/appsignal/appsignal-ruby/issues/1606.

`Appsignal::Rack::GrapeMiddleware` builds the transaction action from
the HTTP method, path and API class on the Grape endpoint's `options`
Hash. Grape 4 keeps those three in a value object behind a protected
reader, so all three lookups return nil and every request through a
Grape API raises `NoMethodError`. The middleware now reads them from
the endpoint's route on Grape 4, and keeps reading `options` on Grape 3
and older, because the route spells the path as the full route template
and switching to it would rename their actions.

### DIFF
--- a/.changesets/support-grape-4.md
+++ b/.changesets/support-grape-4.md
@@ -1,0 +1,8 @@
+---
+bump: patch
+type: add
+---
+
+Add support for Grape 4. On Grape 4, every request through a Grape API raised a `NoMethodError`. Applications on Grape 3 and below were not affected.
+
+On Grape 4, the action name and the reported path now describe the endpoint's full route, so they include the API prefix, the path version and the mount point. They also no longer end in a trailing slash when the endpoint declares no path of its own, so an endpoint reported as `GET::My::Api#/users/:id/` is now reported as `GET::My::Api#/users/:id`. Action names on Grape 3 and below do not change.
--- a/lib/appsignal/rack/grape_middleware.rb
+++ b/lib/appsignal/rack/grape_middleware.rb
@@ -16,26 +16,56 @@ module Appsignal
 
       def add_transaction_metadata_after(transaction, request)
         endpoint = request.env["api.endpoint"]
-        unless endpoint&.options
+        request_method, klass, path = endpoint && endpoint_action(endpoint)
+        unless path
           super
           return
         end
-
-        options = endpoint.options
-        request_method = options[:method].first.to_s.upcase
-        klass = options[:for]
-        namespace = endpoint.namespace
-        namespace = "" if namespace == "/"
-
-        path = options[:path].first.to_s
-        path = "/#{path}" if path[0] != "/"
-        path = "#{namespace}#{path}"
 
         transaction.set_action_if_nil("#{request_method}::#{klass}##{path}")
 
         super
 
         transaction.set_metadata("path", path)
+      end
+
+      # Returns the HTTP method, API class and path that make up the action
+      # name, or nil when the endpoint does not describe a route.
+      def endpoint_action(endpoint)
+        options = endpoint.options
+        return unless options
+
+        if options.key?(:path)
+          # Only Grape 3 and older populate `options[:path]`. Their route is
+          # readable too, but its path is the full route template, so reading
+          # it would rename the actions these applications already report.
+          endpoint_action_from_options(endpoint, options)
+        else
+          # Grape 4 keeps these three in a value object behind a protected
+          # reader, leaving the route as the only public source.
+          endpoint_action_from_route(endpoint)
+        end
+      end
+
+      def endpoint_action_from_options(endpoint, options)
+        namespace = endpoint.namespace
+        namespace = "" if namespace == "/"
+
+        path = options[:path].first.to_s
+        path = "/#{path}" if path[0] != "/"
+
+        [
+          options[:method].first.to_s.upcase,
+          options[:for],
+          "#{namespace}#{path}"
+        ]
+      end
+
+      def endpoint_action_from_route(endpoint)
+        route = endpoint.routes.first
+        return unless route
+
+        [route.request_method.to_s.upcase, endpoint.api, route.origin]
       end
     end
   end
--- a/spec/lib/appsignal/rack/grape_middleware_spec.rb
+++ b/spec/lib/appsignal/rack/grape_middleware_spec.rb
@@ -26,6 +26,30 @@ if DependencyHelper.grape_present?
       end.to raise_error(exception_class, exception_message)
     end
 
+    let(:expected_method) { "GET" }
+
+    # Where the two Grape majors report a route differently, each gets its own
+    # "in Grape 3" and "in Grape 4" examples rather than one example with a
+    # version-dependent expectation. The spec coverage audit works per source
+    # line and asks only that some build matrix combination runs each one, so
+    # examples shared between the two majors would let a Grape 4 run stand in
+    # for a Grape 3 one that the matrix had stopped running. These helpers hold
+    # the assertions so that only the expected path is written out twice.
+    def expect_reported_route_in_agent_mode
+      expect(last_transaction).to have_action(expected_action)
+      expect(last_transaction).to include_metadata(
+        "path" => expected_path,
+        "method" => expected_method
+      )
+    end
+
+    def expect_reported_route_in_collector_mode
+      expect(root_span.name).to eq(expected_action)
+      expect(root_span.attributes["appsignal.action_name"]).to eq(expected_action)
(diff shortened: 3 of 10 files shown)


.changesets/support-grape-4.md
```diff
@@ -0,0 +1,8 @@
+---
+bump: patch
+type: add
+---
+
+Add support for grape 4. On grape 4, every request through a Grape API raised a `NoMethodError`. Applications on grape 3 and below were not affected.
```
unflxw: Let's capitalise it consistently to either "`grape`" (the package) or "Grape".

lib/appsignal/rack/grape_middleware.rb
```diff
@@ -16,26 +16,58 @@ def initialize(app, options = {})
...
+
+      # Returns the HTTP method, API class and path that make up the action
+      # name, or nil when the endpoint does not describe a route.
+      #
+      # grape 4 moved the HTTP method, path and API class off the endpoint's
+      # `options` Hash and into a value object behind a protected reader, so
+      # the endpoint's route is read instead. Only grape 4 and newer take that
+      # branch. The route spells the path as the full route template, which
+      # includes the API prefix, the path version and the mount point, so
+      # reading it on grape 3 and older would rename their actions.
```
unflxw: Let's move these comments to the branches they describe inside the function, where they're in context, shortening them in the process.

unflxw:
You'll notice that the way routes are reported changes between Grape 3 and Grape 4. That is necessary, since Grape 4 doesn't allow us to fetch the route in the same way that Grape 3 did, but it's also, in my view, an improvement, since the new route description is more informative while still being high-cardinality.

We _could_ change how routes are reported in Grape 3 to match how they're now reported in Grape 4, but then that would change the performance incident groupings of customers using Grape 3 on a patch/minor AppSignal version bump. It makes more sense to me that they would change when upgrading to Grape 4 instead, though that is ultimately a subjective feeling of mine and not a strongly held opinion.

.changesets/support-grape-4.md
unflxw: Normalized to "Grape" for the library in prose. That is what the changelog already mostly uses, and it matches the Grape docs page. Lowercase `grape` is now only the package name, which leaves it in the gemfiles and the gemset names.

I applied the same thing beyond the changeset, so the commit title, the commit body and the spec comments now read "Grape" too. The `grape-3.gemfile` comment stays lowercase, because there it really is the package and its version constraint.

lib/appsignal/rack/grape_middleware.rb
unflxw: Moved and shortened. The `if` branch now says why Grape 3 and older keep reading `options`, and the `else` branch says why Grape 4 leaves the route as the only public source. The method keeps just the one line about what it returns.

unflxw:
Agreed, and that is the reasoning the code now carries: the `if` branch explains that Grape 3 and older keep reading `options` so their action names do not change, and the `else` branch explains that Grape 4 leaves the route as the only public source.

Both spellings of the divergence are covered by tests. `DependencyHelper.grape4_present?` picks the expected value in three places: the dropped trailing slash, the API prefix with the path version, and the mount point. I checked they are load-bearing by inverting all three, which fails three examples on the `grape-3` gemset and three on `grape-4`.

I also documented it for customers in appsignal/appsignal-docs#246, which adds a short "Grape 4" section to the Grape page. That says the groupings change on the Grape upgrade rather than on an AppSignal one.

spec/lib/appsignal/rack/grape_middleware_spec.rb
```diff
@@ -156,6 +156,14 @@ def perform
         Rack::MockRequest.env_for("/users/123", :method => "GET")
       end
 
+      # The endpoint declares no path of its own, so Grape 3 and older report
+      # the namespace with the endpoint's default "/" path appended. Grape 4
+      # reports the route's own template, which has no trailing slash.
+      let(:expected_path) do
+        DependencyHelper.grape4_present? ? "/users/:id" : "/users/:id/"
```
unflxw: I'm thinking that, rather than doing this per-version expectation change within the test, we should do something like defining separate "in Grape 4" and "in Grape 3" `describe` blocks that run the same tests with different `let(:expected_path)` values.

The purpose of this would be for [our check that every test runs in CI](https://github.com/appsignal/appsignal-ruby/pull/1598) to flag it if we ever accidentally stop running Grape 3 tests.

lipskis (approved):

spec/lib/appsignal/rack/grape_middleware_spec.rb
unflxw: Done, and the audit does catch it now.

Each of the three version-dependent contexts has an `in Grape 3` and an `in Grape 4` `describe`, gated with `:if => DependencyHelper.grape4_present?`, differing only in `let(:expected_path)`. The assertions moved into two helper methods so that only the expected path is written out twice.

One thing worth knowing, because it decides the shape. The examples cannot be shared between the two describes. The audit counts distinct examples per source line and asks only that some combination ran each line at least once, so a single `it` included from both majors would still be covered by the Grape 4 run alone, and dropping Grape 3 would stay invisible. That is the same reason the version ternary did not work. Each major therefore needs its own `it` calls at its own lines, which is what makes the miss visible.

I checked it end to end rather than assuming, by running `script/audit_spec_coverage.rb` against example lists from the four grape gemsets and a cut-down workflow. With both majors in the matrix it reports nothing for this file. With the Grape 3 gemsets removed it reports exactly the six Grape 3 examples:

```
spec/lib/appsignal/rack/grape_middleware_spec.rb:194  it "sets non-unique route_param path in agent mode", :agent_mode do
spec/lib/appsignal/rack/grape_middleware_spec.rb:201  it "sets non-unique route_param path in collector mode", :collector_mode do
spec/lib/appsignal/rack/grape_middleware_spec.rb:352  it "sets the prefixed and versioned path in agent mode", :agent_mode do
spec/lib/appsignal/rack/grape_middleware_spec.rb:359  it "sets the prefixed and versioned path in collector mode", :collector_mode do
spec/lib/appsignal/rack/grape_middleware_spec.rb:421  it "sets the mounted path in agent mode", :agent_mode do
spec/lib/appsignal/rack/grape_middleware_spec.rb:428  it "sets the mounted path in collector mode", :collector_mode do
```

I also confirmed `:if => false` excludes rather than skips, so these never show up as pending everywhere and trip the other check instead.

Backported to #1611, which has one example per major since that spec is agent mode only. The audit does not exist on `4.x-stable`, so there the split buys consistency rather than a check.

PR #1615, opened 2026-09-14
Update agent to version 0.37.2

AppSignalReleaseBot:
The agent update and changesets are updated automatically.

[skip review]


unflxw:
@tombruijn Should we merge this? If so, we should cherry-pick it onto `4.x-stable` as well.

tombruijn:
@unflxw I'm okay with leaving this open for a little while longer until we hear back from the customer or I have looked into the issue a little more.

PR #1616, opened 2026-09-16
Correct an array option set to a YAML null

unflxw:
An array option written as a null in the config file holds `nil`, which nothing that reads one expects. The options that reach the agent are joined into a string when AppSignal starts, and `filter_metadata` and `filter_session_data` are searched while a transaction is sampled, so both raise a `NoMethodError`.

Correct them to the empty list the option would have held. This runs with the other overrides, after the values worked out from other options have been merged, so a `nil` that reached one of those is corrected too.


unflxw:
Folded back into #1609, which is where the specs for it can name the collector-mode allowlists.

Issue #1617, opened 2026-09-21
`Appsignal.active?` returns `false` inside `config/initializers` on Rails 8.1


inkstak:
## Environment

- AppSignal gem: reproduced on both 4.8.6 and 4.10.3 (latest)
- Rails: 8.1.3.1
- Ruby: 4.0.5 (aarch64-linux)

Appsignal gem version was unchanged across our Rails 8.0 → 8.1 upgrade.  
Only the Rails version changed, so this looks like a Rails 8.1-side ordering shift rather than a regression in the AppSignal gem.

## Describe the bug

Some of our own app code checks `Appsignal.active?` inside initializers to conditionally configure logging.
Example:

```ruby
# config/initializers/lograge.rb

Rails.application.configure do
  if Appsignal.active?
    config.lograge.logger = Appsignal::Logger.new(
      "rails",
      format: Appsignal::Logger::LOGFMT
    )
  end
end
```

```ruby
# config/initializers/sidekiq.rb

Sidekiq.configure_server do |config|
  if Appsignal.active?
    config.logger = Appsignal::Logger.new("sidekiq")
    config.logger.broadcast_to(Sidekiq::Logger.new($stdout))
  end
end
```

Under Rails 8.0 this correctly reflected whether Appsignal was going to end up active.  
After upgrading to Rails 8.1, `Appsignal.active?` is always false at that point, even in environments where Appsignal does become active moments later in boot.

I want to flag this even though your own docs don't rely on `.active?` for the Lograge example, because `Appsignal.active?` is public API, and this ordering change silently breaks any app code that reasonably assumes it reflects final state once regular initializers have started running.

### Root cause (as far as we can tell)

`Appsignal::Integrations::Railtie` registers its startup logic as a plain `initializer "appsignal.configure_rails_initialization"` ([railtie.rb#L17](https://github.com/appsignal/appsignal-ruby/blob/v4.10.3/lib/appsignal/integrations/railtie.rb#L17)) with no `before:`/`after:` ordering constraint relative to Rails' own `load_config_initializers` initializer (which is what actually runs `config/initializers/*.rb`). 

## Expected behavior

Either:
1. `appsignal.configure_rails_initialization` runs before `load_config_initializers`, so `Appsignal.active?` reflects final state by the time app initializers run (matches Rails 8.0 behavior on our end), or

[...]

PR #1618, opened 2026-09-21
Start AppSignal before Rails initializers

unflxw:
Fixes https://github.com/appsignal/appsignal-ruby/issues/1617.

The `appsignal.configure_rails_initialization` initializer declares no
ordering constraint, so Rails sorts it at its index in the initializer
collection, and gem load order decides that index. A plain
`Rails::Railtie` that loads before `appsignal` and declares
`after: :load_config_initializers` pulls the engine chain, and with it
the app's `config/initializers`, ahead of AppSignal's start. In those
apps `Appsignal.active?` returns `false` inside an initializer, errors
raised there go unreported, and `config/appsignal.rb` loses to any
`Appsignal.configure` call in an initializer. Declaring
`before: :load_config_initializers` pins the initializer ahead of the
app's own, whichever gems are loaded first.

Forward-ported to `main` in #1619.


unflxw:
@tombruijn Here's one where I'd appreciate a review from you specifically, as you're more familiar with our initializer system and the `start_at` toggle.

The issue is that in certain scenarios, other gem's Railties can cause our own gem's initializer to happen _after_ every `config/initializer/*.rb` (which are ran during `load_config_initializers`). This is partly dependent on gem load order, which for applications using `Bundler.require!` is equivalent to the `Gemfile`'s declaration order.

If those config initializers then check `Appsignal.active?` to conditionally apply AppSignal-specific configuration, the answer be `false` when that happens, even though AppSignal will still be initialized immediately after.

bjacquet (approved):

lipskis (approved):

tombruijn (approved):

